PR #2, opened 2021-11-16
Add CI

pezholio:
This adds a Github action to run our tests in CI. Eventually, we'll probably a Docker image to run the tests in, but this is a start.


script/ci
```diff
@@ -0,0 +1,24 @@
...
+
+set -e
+
+export DATABASE_URL="postgres://$POSTGRES_USER:$POSTGRES_PASSWORD@$POSTGRES_HOST:$POSTGRES_PORT/$POSTGRES_DB"
+
+cd "$(dirname "$0")/.."
+
+echo "==> Installing dependencies..."
+
+npm install
```
Gweaton, line 14: Do we want to use `npm ci` here?

> This command is similar to npm install, except it's meant to be used in automated environments such as test platforms, continuous integration, and deployment -- or any situation where you want to make sure you're doing a clean install of your dependencies.

Taken from: https://docs.npmjs.com/cli/v8/commands/npm-ci

script/ci
```diff
@@ -0,0 +1,24 @@
...
+
+echo "==> Running database migrations"
+
+npm run typeorm migration:run
+
+echo "==> Building assets..."
+
+npm run build:assets
```
Gweaton, line 22: Do we need to build the NestJS app here too with `npm run build`?

Gweaton (commented):
Nice work, just a couple of comments

script/ci
pezholio: `npm run start:test` will build the app anyway. We'll probably want to do a production build for deployment though, but that can come later and in a different pipeline.

Gweaton (approved):

script/ci
Gweaton: Yeah, I only found out about it just now looking at the [starter workflows](https://github.com/actions/starter-workflows/blob/main/ci/node.js.yml)!

PR #3, opened 2021-11-16
Add an auto rebase action

pezholio:
We have the repo set up to make sure branches are up to date before merging. While Github has a button to update a branch, this adds a merge commit, which makes the history messy.

This PR adds a Github action that listens for a comment with `/rebase` in the body and then automatically rebases against the target branch.

It also automatically rebases if the branch is set to automerge, meaning if a branch is behind, but has been approved, you can set it to automerge and let the robots do the work!

The action requires a `AUTO_REBASE_PERSONAL_ACCESS_TOKEN`, which I've already added to the repo.

### DIFF
--- a/.github/workflows/automatic-rebase.yml
+++ b/.github/workflows/automatic-rebase.yml
@@ -1,0 +1,77 @@
+on:
+  issue_comment:
+    types:
+      - created
+  pull_request_target:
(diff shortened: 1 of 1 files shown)


Gweaton (approved):
Nice, great idea!

PR #6, opened 2021-11-18
Add Profession and Organisation data modelling decisions as an ADR

Gweaton:
This includes a PlantUML diagram to help visualise the models.

This is our first pass at designing our data structures, based on data taken from the original EU database of Regulated Professions, and will change over time, particularly as there are a few fields we know we're going to need to add in future. 

We probably won't need to add a new ADR every time this changes, but it makes sense to keep this decision documented in our codebase to document our initial assumptions.


doc/architecture/decisions/0009-represent-data-structures-in-a-relational-database.md
```diff
@@ -0,0 +1,43 @@
...
+
+Having this starting data structure will allow us to begin building the functionality for allowing administrators to create new Organsations and Professions in the service, and for members of the public to view the relvant information from this starting set of fields.
+
+This is by no means a finalised data structure, as we're aware of the key fields missing, as mentioned above.
+
+We'll need to further explore these fields in user research and in discussions with the wider BEIS team to determine the best structure and use case for them.
+
+### Questions / things to consider
+
+- For professions in the same field, with the same nature of work that might belong to a different jurisdiction or region e.g Primary School Teacher - **Northern Ireland** and Primary School Teacher - **Scotland**, should we represent them as the same `Profession` internally but link them to their respective regions / organisations, or do we require them to be separate professions?
```
pezholio, line 40: In the EU database, it looks like there's seperate sub-professions for the different constituent countries where appropriate, we might be able to be a bit smarter than this, but I think what we have is fine for now

doc/architecture/diagrams/organisations-and-professions/organisations-and-professions.wsd
```diff
@@ -0,0 +1,41 @@
...
+  * name: string
+  alternateName: string
+  * description: string
+  * occupationLocation: string[]
+  * regulationType: string
+  * qualificationLevel: string
+  reservedActivities: string[]
+}
+
+entity Legislation {
```
pezholio, line 27: These should probably have IDs

doc/architecture/diagrams/organisations-and-professions/organisations-and-professions.wsd
```diff
@@ -0,0 +1,41 @@
...
+  name: string
+  url: string
+}
+
+entity OccupationalCategory {
```
pezholio, line 32: Can we have an ID for this one too please?

pezholio (commented):
Looks great, only blocker is around IDs. I think we should default to UUIDs for our DB identifiers - makes the URLs less hackable and also doesn't give a potential attacker clues as to how big/small a DB is. I also think they look nice 😃

pezholio (approved):

pezholio:
LGTM 🚢

doc/architecture/decisions/0009-represent-data-structures-in-a-relational-database.md
Floppy: This has been raised with Marianne as a question to explore in research :+1:

Floppy: It might be worth pointing out that if we do want to dedupe these, some sort of Jurisdiction join table between professions and organisations will be needed in future.

Floppy (changes requested):
Looks good. Small suggestion for an improvement around joining professions and organisations.

Floppy (approved):

PR #8, opened 2021-11-18
Setup script improvements: install Node & don't run migrations

Gweaton:
A setup script should install all necessary dependencies. We weren't installing the version of Node here. However, this introduces another dependency on `nodenv`, which we don't install as part of the script - should we?

Running the migrations failed for me running locally, due to not having a `NODE_ENV` variable set at the time of running the script. It runs as part of `npm run start:dev` anyway with the right env variables, so it should be fine to remove here.

### DIFF
--- a/script/setup
+++ b/script/setup
@@ -11,6 +11,11 @@ app_name=$(basename "$PWD")
 dev_db="$app_name-development"
 test_db="$app_name-test"
 
+if ! brew bundle check >/dev/null 2>&1; then
+  echo "==> Installing Homebrew dependencies…"
+  brew bundle install --verbose --no-lock
+fi
+
 echo "==> Dropping and recreating the database..."
 
 dropdb "$dev_db" > /dev/null 2>&1 || true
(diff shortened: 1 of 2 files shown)


script/setup
```diff
@@ -26,6 +26,7 @@ echo "DATABASE_URL=postgres://postgres@localhost:5432/$test_db" > .env.test
 
 echo "==> Installing dependencies..."
 
+nodenv install --skip-existing
```
Gweaton, line 29: Do we want to `brew install nodenv` before this, or add it to the README as a dependency before running this script?

pezholio: Yeah, I think so. Better still, we could add it to a Brewfile and run  `brew bundle install`

Gweaton: Have added this - also with `postgres` just in case the user doesn't have it installed.

pezholio:
LGTM 👍

pezholio (approved):

PR #9, opened 2021-11-19
Use the Datamapper pattern instead of ActiveRecord

pezholio:
Since our last ADR, it's become apparent that the ActiveRecord might not be a good fit for NestJS, so I've written an ADR to reverse that decision and spell out some of the reasons why.

### DIFF
--- a/doc/architecture/decisions/0010-use-the-datamapper-pattern-instead-of-activerecord.md
+++ b/doc/architecture/decisions/0010-use-the-datamapper-pattern-instead-of-activerecord.md
@@ -1,0 +1,118 @@
+# 10. Use the Datamapper pattern instead of ActiveRecord
+
+Date: 2021-11-19
+
+## Status
+
+Accepted
+
+## Summary
+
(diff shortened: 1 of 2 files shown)


doc/architecture/decisions/0010-use-the-datamapper-pattern-instead-of-activerecord.md
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  // Service methods (such as find, where, create, update etc go here)
+}
+```
+
+With this in mind, it makes it much easier to unit test our service
+object without having to access the database directly, making our unit
+tests faster and more reliable, with less set up and tear down to
+worry about. For example:
```
erbridge, line 45: This is the hidden selling point for me. I wonder if this ADR could do with an executive summary?

pezholio: Good plan 👍

Gweaton (approved):

PR #10, opened 2021-11-19
(1896) Create a user model

pezholio:
This creates a basic model to store information about users, who will eventually be logging in to the system. As per #9, we've opted to use the DataMapper pattern over ActiveRecord for ease of testing.

### DIFF
--- a/src/db/migrate/1637317177058-create_users.ts
+++ b/src/db/migrate/1637317177058-create_users.ts
@@ -1,0 +1,40 @@
+import { MigrationInterface, QueryRunner, Table } from 'typeorm';
+
+export class createUsers1637317177058 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    // required for `uuid_generate_v4()`
+    await queryRunner.query(`CREATE EXTENSION IF NOT EXISTS "uuid-ossp"`);
+
+    await queryRunner.createTable(
+      new Table({
+        name: 'users',
+        columns: [
+          {
+            name: 'id',
+            type: 'uuid',
+            isPrimary: true,
+            isUnique: true,
+            default: 'uuid_generate_v4()',
+          },
+          {
+            name: 'name',
+            type: 'varchar',
+          },
+          {
+            name: 'email',
+            type: 'varchar',
+          },
+          {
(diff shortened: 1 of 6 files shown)


src/db/migrate/1637317177058-create_users.ts
```diff
@@ -0,0 +1,37 @@
...
+    await queryRunner.createTable(
+      new Table({
+        name: 'users',
+        columns: [
+          {
+            name: 'id',
+            type: 'uuid',
+            isPrimary: true,
+            isUnique: true,
+            default: 'uuid_generate_v4()',
```
Gweaton, line 14: I'm not sure if you'll need to add this in, but to get `uuid_generate_v4()` to work for me locally, I had to add:
```
    // required for `uuid_generate_v4()`
    await queryRunner.query(`CREATE EXTENSION IF NOT EXISTS "uuid-ossp"`);
```
to the top of the function. I'm adding it in my PR anyway, but it might be worth adding it here if this is the first one going in?

src/users/user.service.spec.ts
```diff
@@ -0,0 +1,56 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { getRepositoryToken } from '@nestjs/typeorm';
+import { Repository } from 'typeorm';
+
+import { User } from './user.entity';
+import { UserService } from './user.service';
+
+const user = new User('email@example.com', 'name', '212121');
```
Gweaton, line 8: Not sure if we need it now, but just in case you're interested, we used [Fishery](https://github.com/thoughtbot/fishery) on the Interventions project, which is based off FactoryBot, in case we ever want to use factories in future.

Gweaton (approved):
Looks good, and a nice example of how to model things in future. Just a couple of optional things I thought I'd mention.

src/db/migrate/1637317177058-create_users.ts
pezholio: Yeah, good plan. Let's do that for this one, and then you can take it out of yours :)

src/users/user.service.spec.ts
pezholio: Oooh, nice!

PR #11, opened 2021-11-19
Add professions and legislations models

Gweaton:
## Changes in this PR

This is a first pass at adding Professions and Legislations as models. It took some time to figure out how TypeORM handles migrations (it doesn't appear to want you to write them unless data changes further down the line in production, which doesn't sit all that well with me.)

I initially tried writing migrations but struggled a bit with getting them working for many-to-many relationships. As mentioned above, it feels like NestJS generates these tables for us with a `sychronize: true` in the DB Config, or by manually running `NODE_ENV=development npm run typeorm schema:sync`. I'd like to investigate this further before we add our entities for `Organisations`.

## What's left to do?

- I'd like to confirm that the many-to-many relationship between is working as we expect it to with the way we've structured our entities.
- Figure out if we should be writing migrations, or if it's ok to rely on `synchronize: true` for now as we add in commit #2d135a0


pezholio:
So, I've managed to get to the bottom of the migrations stuff - we DO want to generate migrations for everything, as using `synchronise: true` in prod could be VERY dangerous. I've updated the documentation about the migrations to add instructions about how we generate migrations from entities, as well as made some tooling quality of life tweaks. I've then gone through all the commits that add/change entities and made sure they have associated migrations.

The one outstanding thing we have is the confidence that the many-to-many relationship is working as we expect. As a day-to-day thing, I think it's fine to just assume it's all working as it should be, but given this is our first rodeo with TypeORM and Nest, it would be nice to confirm this. As NestJS doesn't have a REPL like we'd get in (for example) Rails, we can't as easily just dig in and try some things, so this needs a bit of thought.

pezholio (approved):
I've made a couple of tweaks to this based on my learnings, and I've built out a simple controller and everything works as expected, so I'm happy for this to go in 👍

PR #12, opened 2021-11-19
Run Prettier over everything possible

Gweaton:
This was missing files outside the `src/` and `cypress/` directories.

I've added `.prettierrc` as a symlink of `.gitignore` so we don't run it on things like `node_modules` and it's kept up-to-date with the usual files we'll want to ignore.


webpack.config.js
```diff
@@ -3,56 +3,50 @@ const Encore = require('@symfony/webpack-encore');
...
+    options.sassOptions = {
+      includePaths: ['node_modules/govuk-frontend/'],
+      outputStyle: 'compressed',
+    };
+  });
 
-    // uncomment if you use TypeScript
-    //.enableTypeScriptLoader()
+// uncomment if you use TypeScript
+//.enableTypeScriptLoader()
```
erbridge, line 39: Unrelated to this, but should this be uncommented?

pezholio:
This failed initially, because `.gitignore` was in `.dockerignore`, but I've now fixed that 🙌

pezholio (approved):

PR #13, opened 2021-11-22
(1901) Integrate with Okta for user login

pezholio:
This uses the Okta Open ID Middleware (https://github.com/okta/okta-oidc-middleware) to integrate our app with Okta and
add simple login / logout functionality. It's not (yet) linked to a real user in our application, but this can come later.


src/authentication/authentication.guard.ts
```diff
@@ -0,0 +1,32 @@
...
+ * it is true or false.
+ *
+ * @example (in a controller, as metadata above a controller action)
+ *
+ * @UseGuards(AuthenticationGuard)
+ *
+ */
+@Injectable()
+export class AuthenticationGuard implements CanActivate {
+  constructor(private reflector: Reflector) {}
```
bryangaledxw, line 22: Do we need reflector here?

pezholio: Ah, no we don't. I think I cargo culted that from somewhere else. Will fix 😃

src/app.controller.ts
```diff
@@ -10,4 +20,18 @@ export class AppController {
...
+  @Render('admin')
+  admin(): object {
+    return {};
+  }
+
+  @Get('/logout')
+  @UseGuards(AuthenticationGuard)
+  @Redirect('/')
+  logout(@Req() req: Request): object {
+    return req.logout();
```
bryangaledxw, line 35: This doesn't seem to perform a logout in the "user has to re-enter their password with Okta" sense. Is there some equivalent to this we want - https://github.com/panva/node-openid-client/blob/main/docs/README.md#clientendsessionurlparameters - a URL to redirect the user to that will perform an Okta-side logout?

bryangaledxw: Hmm, the more I play with this the more it doesn't seem like a reliable logout? If I visit logout, then admin, then admin a second time, I'm logged in

bryangaledxw: Ah n/m, that's the redirect to login doing its thing

src/main.ts
```diff
@@ -50,6 +61,8 @@ async function bootstrap() {
     }),
   );
 
+  app.useGlobalFilters(new UnauthorizedExceptionFilter());
+
   await app.listen(3000);
```
bryangaledxw, line 66: Going by the docs, we want something like:
`await new Promise((res, rej) => {oidc.on('ready', res); oidc.on('error', rej);}); `

before starting the server

src/app.controller.ts
bryangaledxw: Ah k, see - https://github.com/okta/okta-oidc-middleware#oidcforcelogoutandrevoke

I think this is what we want instead? Though note the caution on logout on GET. Though also, the middleware provides a built in POST 'logout' route

pezholio: Ha, nice. I think you're right though - we should really be `POST`ing for a logout

src/main.ts
pezholio: Good catch. I think this is what might have been causing those errors we were seeing too.

cypress/integration/application.spec.js
```diff
@@ -3,4 +3,22 @@ describe('/', () => {
...
+
+      cy.get('body').should('contain', 'Hello Admin!');
+    });
+
+    it('allows me to logout', () => {
+      cy.visit('/admin');
+
+      cy.get('[data-purpose=logout]').click();
+    });
+  });
```
bryangaledxw, line 23: Should we test that admin redirects when if accessed when not logged in, and that logout does actually log the user out?

cypress/support/commands.js
```diff
@@ -1,25 +1,49 @@
...
-// Cypress.Commands.add('drag', { prevSubject: 'element'}, (subject, options) => { ... })
-//
-//
-// -- This is a dual command --
-// Cypress.Commands.add('dismiss', { prevSubject: 'optional'}, (subject, options) => { ... })
-//
-//
-// -- This will overwrite an existing command --
-// Cypress.Commands.overwrite('visit', (originalFn, url, options) => { ... })
+Cypress.Commands.add('login', () => {
```
bryangaledxw, line 1: So I understand why we need this, is this so we're not depending relying on the HTML of the Okta login page not changing? Am I right in thinking Cypress _could_ just populate the Okta login page with a username and password?

src/app.controller.ts
```diff
@@ -10,4 +12,11 @@ export class AppController {
   getHello(): object {
     return {};
   }
+
+  @Get('/admin')
+  @UseGuards(AuthenticationGuard)
```
bryangaledxw, line 17: Again for my understanding, I guess down the line we want a guard that checks not just that the user is logged in, but that they actually have admin permissions. Or do you see that being logic in the function body?

src/main.ts
```diff
@@ -42,6 +44,15 @@ async function bootstrap() {
   app.setBaseViewsDir(views);
   app.setViewEngine('njk');
 
+  app.use(
+    require('express-session')({
```
bryangaledxw, line 48: Can we install @types/express-session and do `import * as session from 'express-session';` to keep things Typescript-y?

views/admin.njk
```diff
@@ -0,0 +1,26 @@
+{% extends "base.njk" %}
+
+{% block header %}
+  {{
+    govukHeader({
+      navigation: [
+          {
+            html: '<form method="post" action="/logout">
```
bryangaledxw, line 8: Is it worth adding logout to the standard header, not just for the admin page (only shown if logged in) as part of this PR?

src/main.ts
bryangaledxw: Also, should we have a ticket somewhere to choose a production express-session store? See note about MemoryStore - https://github.com/expressjs/session

cypress/support/commands.js
pezholio: Yeah, it's considered best practice in Cypress land to not test anything outside of our domain. If we wanted to be _really_ clever, we could set up our own OAuth server, or stub out all the calls that we expect our server to make, but this seems likie a good half way house, especially now we can store and reuse the session at will throughout the suite.

src/app.controller.ts
pezholio: Yeah, I think we'll have that as a guard too. We can set metadata to tell the guard what permissions a user needs too. There's some docs on this use case here https://docs.nestjs.com/guards#role-based-authentication

src/main.ts
pezholio: Both excellent shouts - I'm new to this world, so anything like this that you spot, let me know 👍

views/admin.njk
pezholio: I'm leaning towards no for this. I'm seeing the two sections of the site as distinct and seperate, so having seperate layouts gives us flexibility. I could be wrong though!

pezholio:
Closing this in favour of building something a tad more generic.

PR #14, opened 2021-11-23
Create generic user

bryangaledxw:
This demonstrates a multi-step user creation flow. I've modelled this after the current [PCR test ordering flow](https://www.gov.uk/get-coronavirus-test), including the "edit" query parameter to track if we are editing an already-entered value.

This also adds a Auth0ExternalUserCreationService, and a NullExternalUserCreationService, with the latter being used in testing mode.

Start from `admin/user/create-new-user` to see the flow in action.

@Gweaton adding you on this. Any feedback welcome, but the two places I feel a bit confounded on are
- ~~If there's a nice way to get additional params coming from the session/request body into an error page (see the comment on `personalDetailsPost`)~~
- ~~How I go about setting up a transaction in `confirmPost` that eliminates the possible race condition noted in the comment~~

Any thoughts on those would be appreciated


src/app.controller.ts
```diff
@@ -1,13 +1,37 @@
...
+
+  @Post('/register')
+  @Render('register')
+  async registerPost(@Body() body): Promise<object> {
+    const name = body['name'];
+    const email = body['email'];
+
+    const identifier = await this.appService.createAuth0User(email);
+
+    await this.userService.add(new User(email, name, identifier));
```
pezholio, line 33: What happens if we create an Auth0 user, but the user creation fails? Or vice versa? It might be best to wrap it all in a transaction and:

* Part create the user with name and email
* Create the Auth0 user
* If the Auth0 call fails, rollback the transaction
* If the Auth0 call succeeds, add the identifier and save the User

src/app.controller.ts
```diff
@@ -1,13 +1,37 @@
...
+
+  @Get('/register')
+  @Render('register')
+  registerGet(): object {
+    return {};
+  }
+
+  @Post('/register')
+  @Render('register')
+  async registerPost(@Body() body): Promise<object> {
```
pezholio, line 27: Can we use a  [DTO](https://docs.nestjs.com/controllers#request-payloads) here? That way we can define the shape of the data we expect, and we get an error presented back to the user if something is wrong / missing.

src/app.controller.ts
pezholio: We'll also probably want to raise an error and rescue from it to present an error back to the user if the API call fails, in a similar way to [how we handle validation errors](https://github.com/UKGovernmentBEIS/regulated-professions-register/blob/main/src/validation/validation-exception.filter.ts)

views/register.njk
```diff
@@ -0,0 +1,31 @@
+{% extends "base.njk" %}
+{% from "govuk/components/input/macro.njk" import govukInput %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+
+{% block content %}
```
pezholio, line 6: It would be good if we could nick the approach to showing errors in my Nest example project here https://github.com/pezholio/nest-example/blob/main/views/shared/_errors.njk, so we can use the partial like so:

```
{% include "../shared/_errors.njk" %}
```

src/app.controller.ts
bryangaledxw: Where would user creation fail here? On the db write? In that case, I think it's always going to be possible to wind up with an Auth0 user being created, but no user being created in our DB

I do have in mind though that "email already exists in Auth0 but not in our db" is a case we'll need to handle

src/registration/registration.controller.ts
```diff
@@ -0,0 +1,176 @@
...
+    private readonly externalUserCreationService: ExternalUserCreationService,
+  ) {}
+
+  @Get('/start')
+  @Render('registration/start')
+  start(): object {
+    return {};
+  }
+
+  @Post('/start')
```
pezholio, line 36: Personally I don't think a multi-page setup is necessary for this kind of thing. This is more a pattern that's used on public-facing services where there's lots of branching logic, or a lot of questions. It's fine to have this on one page, and makes all this a lot simpler.

pezholio: FWIW - this is the pattern we used on the last project I was on:

http://g.recordit.co/sUNUYrLTnR.gif

pezholio: I think I'd prefer to have a flash message a la https://design-system.service.gov.uk/components/notification-banner/#reacting-to-something-the-user-has-done rather than the confirmation page in the example above

src/registration/null-user-creation-service.ts
```diff
@@ -0,0 +1,11 @@
+import { randomUUID } from 'crypto';
+import {
+  CreateExternalUserResult,
+  ExternalUserCreationService,
+} from './external-user-creation.service';
+
+export class NullUserCreationService extends ExternalUserCreationService {
```
pezholio, line 7: Oh, this is smart - is this for use in the e2e tests?

src/registration/registration.controller.ts
```diff
@@ -0,0 +1,176 @@
...
+import { QueryDto } from './dto/query.dto';
+import { RegisterEmailDto } from './dto/register-email.dto';
+import { RegisterNameDto } from './dto/register-name.dto';
+import { RegisterSessionDto } from './dto/register-session.dto';
+import { ExternalUserCreationService } from './external-user-creation.service';
+
+const SESSION_KEY = 'registration';
+const TOTAL_STEPS = 2;
+
+@Controller('/register')
```
pezholio, line 23: I'd prefer a RESTful approach to naming things here, so we'd have the controller called `/user` and then have `new`, `create`, `edit`, `update` controller methods.

src/users/user.entity.ts
```diff
@@ -11,7 +11,7 @@ export class User {
   @Column()
   name: string;
 
-  @Column()
+  @Column({ unique: true }) // Should I be surprised this change doesn't generate a migration?
```
pezholio, line 14: Hmmm - that's odd. Is that after running `npm run typeorm:migration:generate`?

pezholio:
Cool - all this is looking on the right track. Thanks! I'd be more comforable with more unit and e2e tests, but I guess those are coming.

One point - we tend to try and break down our commits in small chunks to make reviewing easier. If you could do that, that would be great! 👍

src/registration/null-user-creation-service.ts
bryangaledxw: Yep, so this will get used when `process.env.NODE_ENV == 'test'`

src/registration/registration.controller.ts
bryangaledxw: I think I need more hand-holding on this one. Unsure how that maps to the current flow and how e get nice user-facing URLs with that

src/registration/registration.controller.ts
bryangaledxw: Ok to to resolve this following our chat with Ming?

src/app.controller.ts
bryangaledxw: I think this is resolved in the latest push?

views/registration/done.njk
```diff
@@ -0,0 +1,34 @@
...
+<div class="govuk-width-container">
+
+  <main class="govuk-main-wrapper">
+
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+
+        {{ govukPanel({
+          titleText: "User added",
+          html: "User <strong>" + email + "</strong> has been added"
```
bryangaledxw, line 15: I very strongly suspect this is not the right way to do this, though couldn't find the correct way

src/users/personal-details.controller.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    if (!validator.valid()) {
+      res.render('user/personal-details', {
+        edit,
+        name: personalDetailsDto.name,
+        email: personalDetailsDto.email,
+        errors: new ValidationFailedError(validator.errors).fullMessages(),
+      });
+      return;
+    }
```
pezholio, line 83: You shouldn't need to do this here, if you add a decorator, Nest should pick up any validation errors and pass them back to the template.

```
@UseFilters(new ValidationExceptionFilter('user/personal-details', 'personalDetailsDto'))
```

That said, if there's invalid data, then Nest _should_ be raising an error.

src/users/personal-details.controller.ts
```diff
@@ -0,0 +1,109 @@
...
+  UserCreationFlowSession,
+  UserCreationFlowStep,
+} from './helpers/user-creation-flow-session.helper';
+
+@Controller('/user/personal-details')
+export class PersonalDetailsController {
+  constructor(private readonly userService: UserService) {}
+
+  @Get()
+  @Render('user/personal-details')
```
pezholio, line 27: Can we have views in the format `$MODULE/$CONTROLLER/$ACTION` please? So, this should be `user/personal-details/new`

src/users/auth0-user-creation.service.ts
```diff
@@ -0,0 +1,41 @@
+import { ManagementClient } from 'auth0';
+import { randomUUID } from 'crypto';
+import {
+  CreateExternalUserResult,
+  ExternalUserCreationService,
+} from './external-user-creation.service';
+
+export class Auth0UserCreationService extends ExternalUserCreationService {
```
pezholio, line 8: Can we have any new classes (outside of the usual module, controller, service etc classes) documented in a [JSDoc](https://jsdoc.app/) stylee please? Just got half an eye on maintainability going forward

src/users/auth0-user-creation.service.ts
```diff
@@ -0,0 +1,41 @@
...
+  public async createExternalUser(
+    email: string,
+  ): Promise<CreateExternalUserResult> {
+    const url = process.env.AUTH0_DOMAIN;
+    const domain = url.startsWith('https://') ? url.slice(8) : url;
```
pezholio, line 13: I've handled this in #24 by adding a `AUTH0_HOSTNAME` env var. That said, I do wonder if it might be better if we just strip the protocol from our env vars and have the domain. I kept the domain in originally, because I had half a mind to mock it out in Cypress at some point, but I think that's going to be too much of a faff. I'm quite happy with the puppeteer approach we have now.

src/users/auth0-user-creation.service.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    // Since we only support password authentication, assume we either get zero
+    // or one users returned
+    if (users.length > 0) {
+      return { result: 'user-exists', externalIdentifier: users[0].user_id };
+    }
+
+    const user = await client.createUser({
+      email,
+      password: randomUUID(),
```
pezholio, line 35: I've just tried this locally and I don't get a prompt to change my password. Is there anyway we can set up a user without a password and send them a password reset email?

src/users/personal-details.controller.ts
pezholio: D'oh, I missed this comment:

```
    // Intentially don't use `ValidationExceptionFilter`, as we have additional
    // parameters to get into our template
```

src/users/personal-details.controller.ts
```diff
@@ -0,0 +1,109 @@
...
+        errors: new ValidationFailedError(validator.errors).fullMessages(),
+      });
+      return;
+    }
+
+    // Don't talk to Auth0 yet, but at least check our own DB
+    if (await this.userService.findByEmail(personalDetailsDto.email)) {
+      const errors = {
+        email: { text: 'A user with this email address already exists' },
+      };
```
pezholio, line 89: Can we add these to an errors object so we only have to call `res.render` once, or at maybe put the call to `res.render` in a function to DRY this up a bit?

src/users/user.controller.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  @Get('/done')
+  @Render('user/done')
+  done(@Session() session): object {
+    const userCreationFlowSession = new UserCreationFlowSession(
+      session,
+      UserCreationFlowStep.Complete,
+    );
+    const { email } = userCreationFlowSession.sessionDto;
```
pezholio, line 94: Should we clear our session var at this point too? I've just tried this out locally, and going back in populates the form with the last user I added

src/users/personal-details.controller.ts
bryangaledxw: tbh I suspect the case where all we need to render the template is the errors is going to be fairly rare. I'm also wondering how we handle things if/when we want to improve the error messages displayed to the user. Right now we're relying on the exception message being good enough

src/users/personal-details.controller.ts
```diff
@@ -0,0 +1,109 @@
...
+import { ValidationFailedError } from '../validation/validation-failed.error';
+import { Validator } from '../helpers/validator';
+import { UserService } from './user.service';
+import { PersonalDetailsDto } from './dto/personal-details.dto';
+import {
+  UserCreationFlowSession,
+  UserCreationFlowStep,
+} from './helpers/user-creation-flow-session.helper';
+
+@Controller('/user/personal-details')
```
pezholio, line 22: Can we namespace these under `/admin` please?

pezholio: We can also add `@UseGuards(AuthenticationGuard)` to only allow logged in users to access this page

src/users/user.controller.ts
bryangaledxw: Will clear session at the end of the confirm step. Feel iffy about clearing session in a GET

src/users/auth0-user-creation.service.ts
bryangaledxw: It didn't seem so at first glance, but will have more of a dig

src/users/user.controller.ts
pezholio: That makes sense. Nice one 👍

src/users/auth0-user-creation.service.ts
pezholio: It might be that we'll have to turn off confirmation emails altogether and send our own with a password reset link. I think that's what we did for RODA. This approach is fine for now though - getting ahead of myself!

src/users/user.controller.ts
bryangaledxw: Gah, no, that doesn't work as we still need the session details on that last page

src/registration/registration.controller.ts
bryangaledxw: ✅

src/users/personal-details.controller.ts
bryangaledxw: ✅

src/users/personal-details.controller.ts
bryangaledxw: ✅

src/users/user.controller.ts
bryangaledxw: Changed this around somewhat - the POST when entering the flow will clear the session, the last step will mark in the session that the user has been created, and once in that state, previous steps will fail _unless_ the user goes through the initial POST again. This tightens things up a bit preventing the user from doing things out-of-order. This seems to be in line with the PCR example (though that has nicer error messages for now)

views/registration/done.njk
bryangaledxw: Seems there isn't an a obvious better alternative - https://github.com/mozilla/nunjucks/issues/899

src/users/personal-details.controller.ts
bryangaledxw: ✅

src/users/auth0-user-creation.service.ts
bryangaledxw: ✅

bryangaledxw:
I believe this should be all feedback addressed now, next step is E2E tests. I want to change instances of `users` to `user` in folder/file names to make things consistent, though would like to leave this as a last step or its own PR due to likelihood of hard-to-resolve conflicts

src/users/helpers/user-creation-flow-session.helper.ts
```diff
@@ -0,0 +1,97 @@
...
+   * Check if the session data is as expected, assuming the provided step of
+   * user creation has been completed
+   *
+   * @param step {UserCreationFlowStep} The step we want to check has been
+   *   completed
+   * @returns {boolean} `true` if the session is valid for a given completed
+   *   step, `false` otherwise
+   */
+  hasCompletedStep(step: UserCreationFlowStep): boolean {
+    if (step == UserCreationFlowStep.Any) {
```
Gweaton, line 61: should this be `===` here? and below, or am I missing something?

src/users/personal-details.controller.spec.ts
```diff
@@ -0,0 +1,157 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { UserService } from './user.service';
+import { User } from './user.entity';
+import { PersonalDetailsController } from './personal-details.controller';
+import { createMock, DeepMocked } from '@golevelup/ts-jest';
+
+const name = 'Example Name';
+const email = 'name@example.com';
+
+describe.only('PersonalDetailsController', () => {
```
Gweaton, line 10: Can we get rid of this `only` before merging into `main`?

src/users/personal-details.controller.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+      const result = controller.new(false, populatedSession);
+
+      expect(result).toEqual({
+        name,
+        email,
+        edit: false,
+      });
+    });
+
+    it('should remaing in edit mode when called in edit mode', () => {
```
Gweaton, line 72: There's a typo here on `remain[g]`

src/users/personal-details.controller.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+      };
+    });
+
+    it('should throw an exception when called in edit mode with an empty session', () => {
+      expect(() => {
+        return controller.create({ name: name, email, edit: 'true' }, {}, res);
+      }).rejects.toThrow();
+    });
+
+    it('should redirct to `confirm` and poulate the session when the email address is not already in use and the body is populated', async () => {
```
Gweaton, line 95: There's a typoe here on `po[p]ulate` (sorry 😅)

src/users/user.controller.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+      };
+    });
+
+    it('should throw an exception when called with an empty session', () => {
+      expect(async () => {
+        await controller.create({}, res);
+      }).rejects.toThrowError();
+    });
+
+    it('should redirct to `done` when the user is successfully created', async () => {
```
Gweaton, line 94: Typo here in `redir[e]ct`

src/users/auth0-user-creation.service.ts
```diff
@@ -0,0 +1,56 @@
...
+export class Auth0UserCreationService extends ExternalUserCreationService {
+  /**
+   * Create a new user in Auth0. The created user will have the provided email
+   * address, and a randonly generated password. In the case where a user
+   * already exists in Auth0 with the provided email address, no new user will
+   * be created, but the Auth0 identifier of the existing user with that email
+   * address will be returned
+   *
+   * @param {string} email The email address of the user to create
+   * @returns {CreateExternalUserResult} The result of the call, containing
```
Gweaton, line 20: The function returns a `Promise<CreateExternalUserResult` below - is that the sort of specificity we've tended to put in our JSDocs? Might be safest to update that here, just to avoid confusion.

Gweaton (commented):
Nice work - I think we're pretty much there! I've added a few pretty small comments (typos etc). 

One thing I would mention, is I found this a bit of a tricky PR to review due to the size of it and some of the commits. Obviously it's a big piece of work that's had a lot of comments to address, so I can understand how it's grown a lot! To make something like this easier to review, I'd probably structure the PR as follows:

- Add new (possibly fairly empty) controller (Admin/UsersController - if renaming, try and amend the initial commit via a rebase, so it looks like you added it with the new name originally, this saves mental overload of remembering something's renamed when reviewing by commit) alongside templates it uses, routing & tests.
- For each new route / page in the journey, add a new commit, alongside tests etc.
- If any complicated setup is required for the above, e.g. validations, helpers, error handling stuff that is too big to squash into the same commit, add this as its own commit with a descriptive commit message saying why it's being added and where it's going to be used.
- I think it's fine to add E2E tests at the end in their own commit, unless it's a small thing in which case probably fine to add alongside the feature code.

Sorry - that's a lot of words, I hope it's helpful. It might be something we can pair on in future as it took a while to get my head around when I was first getting used to working this way. It takes a bit of effort at first, but makes a huge difference when reviewing code.

src/users/auth0-user-creation.service.ts
bryangaledxw: Yep, definitely should be `Promise<>`

src/users/personal-details.controller.spec.ts
bryangaledxw: Yes! This definitely shouldn't be here

src/users/helpers/user-creation-flow-session.helper.ts
bryangaledxw: Nah, should be `===`

cypress/integration/admin/user/create-new-user.spec.ts
```diff
@@ -0,0 +1,70 @@
+describe('/admin/user/create-new-user', () => {
```
Gweaton, line 1: Let's call this descript "Creating a new user" to describe the user journey, rather than the endpoint.

views/user/confirm.njk
```diff
@@ -0,0 +1,103 @@
...
+            {
+              text: "A user with this email address already exists"
+            }
+          ]
+        }) }}
+
+        {% endif %}
+
+        <span class="govuk-caption-xl">Creating RPR user</span>
+        <h1 class="govuk-heading-xl">Check your answers before sending your application</h1>
```
Gweaton, line 36: Can we make this "Check your answers before creating this user"?

views/user/confirm.njk
```diff
@@ -0,0 +1,103 @@
...
+            }
+          ]
+        }) }}
+
+        <h3 class="govuk-heading-s">Now send your application</h3>
+        <p class="govuk-body">By submitting this application you are confirming that, to the best of your knowledge, the details you are providing are correct.</p>
+
+        <form method="post" action="confirm">
+          {{ govukButton({
+            text: "Accept and send",
```
Gweaton, line 85: Let's make this "Create user"

views/user/confirm.njk
```diff
@@ -0,0 +1,103 @@
...
+                    text: "Change",
+                    visuallyHiddenText: "email address"
+                  }
+                ]
+              }
+            }
+          ]
+        }) }}
+
+        <h3 class="govuk-heading-s">Now send your application</h3>
```
Gweaton, line 80: We can probably remove this header and the `<p>` below,  - I think we'll have a content designer in January but for testing, I think "application" might confuse people a bit.

Gweaton (approved):
Just have 2 text updates after testing locally - let's get this one in if you're ok to make the changes?

PR #18, opened 2021-11-26
(1901) User login

pezholio:
This adds a simple integration with Auth0 to allow a registered user to log in. At the moment, we don't actually check a logged in user against the database, but this can come in once #14 is done.


src/app.controller.ts
```diff
@@ -10,4 +11,13 @@ export class AppController {
   getHello(): object {
     return {};
   }
+
+  @Get('/admin')
```
Gweaton, line 15: This might be over the top at this point (and I'm asking due to my lack of experience with Nest JS), but is there a reason we don't put this in an `admin.controller` or something here? Is it just to keep it simple for the time being?

cypress/plugins/auth0.js
```diff
@@ -0,0 +1,80 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+
+const puppeteer = require('puppeteer');
```
Gweaton, line 3: Do these not work as ES6 modules (i.e. can we not `import` stuff)? Is switching to TypeScript for our Cypress tests going to help with that? It'd be good to do that switch sooner, rather than later if possible, before tests start building up.

Gweaton (approved):
This looks like a lot of complex work, well done for figuring how to get everything working with NestJS. I've just added a couple of comments, more questions really.

src/app.controller.ts
pezholio: Yeah, just keeping it simple for now - it's an open question at the mo as to whether we want an admin controller, or we just namespace the relevant create/update actions.

cypress/plugins/auth0.js
pezholio: Yeah, that's on my mental list to do. It's a good shout to keep things consistent.

PR #19, opened 2021-11-29
(1901) Get user from database

pezholio:
This builds on #18 to make sure we have the information about the user from our local database. I've utilised the [`afterCallback`](https://auth0.github.io/express-openid-connect/interfaces/configparams.html#aftercallback) functionality in `express-openid-connect` to leverage this.

To get the tests working here, I've introduced a seed command which uses [nestjs-seeder](https://github.com/edwardanthony/nestjs-seeder) to import the user we expect to have to the test database at runtime. This looks to be a super-useful think we should be able to leverage for a lot of other stuff too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,11 @@
     "test:watch": "jest --watch",
     "test:cov": "jest --coverage",
     "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
-    "test:e2e": "start-server-and-test start:test http://localhost:3000 cy:run"
+    "test:e2e": "start-server-and-test start:test http://localhost:3000 cy:run",
+    "seed": "npm run build && node dist/seeder",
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -46,6 +46,7 @@
     "express-openid-connect": "^2.5.1",
     "file-loader": "^6.2.0",
     "govuk-frontend": "^3.14.0",
+    "jwt-decode": "^3.1.2",
```
Gweaton, line 49: I quite like adding new packages in their own commit messages with an explanation as to why they've been added. If you'd rather introduce it in the same commit, would you mind adding a bit of info as to what we're using `jwt-decode` for to the commit message here (obviously I can scroll through the commit, but it can sometimes be a bit unclear when doing a `git log -S` or something in future.

Gweaton (approved):
Looks good, just one pernickety comment from me.

package.json
pezholio: Cool - I've updated the commit message with a bit more context 👍

PR #20, opened 2021-11-29
(1887) Add Organisation model

Gweaton:
## Changes in this PR

Adds new Organisation model and migrations for it. I'm not 100% sure on which fields should be nullable here, but I've taken a logical guess based on our initial data modelling diagrams.

This PR also switches off the synchronisation behaviour added in #11, as it conflicts with our new method of always running migrations.


pezholio (approved):
LGTM 🚢

PR #21, opened 2021-11-30
Tidy up the Nunjucks config

pezholio:
This adds some type definitions for nunjucks, as well as breaking out all the nunjucks config into one file to make things a bit neater.


bryangaledxw (approved):
Looks good to me

PR #22, opened 2021-11-30
Add an app.json for Heroku

pezholio:
This specifies that we want all review apps to have a Postgres DB set up, and run our seed task to seed the database.

### DIFF
--- a/src/middleware/authentication.middleware.ts
+++ b/src/middleware/authentication.middleware.ts
@@ -8,6 +8,10 @@ import jwt_decode from 'jwt-decode';
 import { UsersModule } from '../users/users.module';
 import { UserService } from '../users/user.service';
 
+const baseURL =
+  process.env['HOST_URL'] ||
+  `https://${process.env['HEROKU_APP_NAME']}.herokuapp.com`;
+
 /**
  * Set up an instance of the `express-openid-connect` authentication middleware for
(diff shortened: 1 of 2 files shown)


bryangaledxw (approved):
I'd missed the seeder work. Looks good!

src/middleware/authentication.middleware.ts
```diff
@@ -33,7 +37,7 @@ export class AuthenticationMidleware {
   public auth(): RequestHandler {
     return auth({
       issuerBaseURL: process.env['AUTH0_DOMAIN'],
-      baseURL: process.env['HOST_URL'],
+      baseURL: baseURL,
```
bryangaledxw, line 40: Sorta outside of this PR, but this is the one place (I think) we're doing:
`process.env['FOO']`, rather than:
`process.env.FOO`

I tend to favour the former when the type system doesn't guarantee that `FOO` will be there, but have been using the latter for consistency with the rest of the codebase. Worth making consistent?

pezholio: Yeah, this is more my naivety than anything! Coming from Ruby I'm used to bracket-based notation for env vars, so I often forget (hence why I did it this way here), but your explanation makes sense. Happy to change it to bracket-based for consistency and type safety 👍

PR #23, opened 2021-11-30
[1919] Move Cypress tests over to TypeScript

Gweaton:
This will allow us the type safety that TypeScript provides, and will
mean we don't have to accommodate both plain JS & TypeScript in the same
application.

There was a fair bit of faff involved in setting up with multiple
`tsconfig.json`s, and I'm not 100% sure this is right yet, but I copied
the setup from Cypress's [Real World
App](https://github.com/cypress-io/cypress-realworld-app).


pezholio (approved):

pezholio:
I've fixed this in George's absence, and got rid of any implicit `any`s in the test code.

PR #24, opened 2021-12-01
Update callback urls for Heroku

pezholio:
This adds a script that gets run after a successful deploy of a Heroku review app to register the callback URLs on Auth0 to allow logins to work properly on Heroku review apps. We include localhost:3000 in this array, because we'll always need this for dev.

### DIFF
--- a/util/set-callback-urls.ts
+++ b/util/set-callback-urls.ts
@@ -1,0 +1,65 @@
+/*
+ * This is a script that gets run after a successful deploy of a Heroku
+ * review app to register the callback URLs on Auth0 to allow logins to
+ * work properly on Heroku review apps.
+ */
+
+import { ManagementClient } from 'auth0';
+import axios from 'axios';
+
+import * as dotenv from 'dotenv';
+
+dotenv.config({ path: '.env.development' });
+
+const callbackUrls = async (): Promise<string[]> => {
+  interface App {
+    pr_number: string;
+  }
+
+  const response = await axios
+    .get<App[]>(
+      `https://api.heroku.com/pipelines/${process.env['HEROKU_PIPELINE_ID']}/review-apps`,
+      {
+        headers: {
+          Accept: 'application/vnd.heroku+json; version=3',
+          Authorization: `Bearer ${process.env['HEROKU_ACCESS_TOKEN']}`,
+        },
+      },
+    )
+    .then((res) => {
+      return res.data;
+    });
+
+  return response.map((app: App) => {
+    return `https://rpr-review-app-pr-${app.pr_number}.herokuapp.com/callback`;
+  });
+};
+
+(async () => {
+  const auth0 = new ManagementClient({
+    domain: process.env['AUTH0_HOSTNAME'],
+    clientId: process.env['AUTH0_CLIENT_ID'],
+    clientSecret: process.env['AUTH0_CLIENT_SECRET'],
+  });
+  const urls = await callbackUrls();
+
+  console.log(
+    'Registering callback URLs for the currently installed Heroku review apps...',
+  );
+
+  auth0.updateClient(
+    { client_id: process.env['AUTH0_CLIENT_ID'] },
+    {
+      callbacks: [...urls, 'http://localhost:3000/callback'],
+    },
+    function (err: Error) {
+      if (err) {
+        console.log(
+          `Registration failed with the following error: '${err.message}''`,
+        );
+      } else {
+        console.log('Success!');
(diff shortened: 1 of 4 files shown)


util/set-callback-urls.ts
```diff
@@ -0,0 +1,55 @@
...
+ * work properly on Heroku review apps.
+ */
+const { ManagementClient } = require('auth0');
+
+import * as axios from 'axios';
+
+require('dotenv').config({ path: '.env.development' });
+
+const callbackUrls = async () => {
+  const response = await axios.default(
```
Gweaton, line 13: I found the `default` here a bit confusing, as I thought it was a type of request method - I can see it's due to the way we import axios. It looks like it's fine to do `import axios from 'axios'`, then we can call `await axios.get()` here which is a bit more readable.

util/set-callback-urls.ts
```diff
@@ -0,0 +1,55 @@
+/*
+ * This is a script that gets run after a successful deploy of a Heroku
+ * review app to register the callback URLs on Auth0 to allow logins to
+ * work properly on Heroku review apps.
+ */
+const { ManagementClient } = require('auth0');
+
+import * as axios from 'axios';
+
+require('dotenv').config({ path: '.env.development' });
```
Gweaton, line 10: Do you think we could try and avoid `require`s in the codebase where possible? Here we can do:
```
import dotenv from 'dotenv';

dotenv.config({ path: '.env.development' });
```
It's a bit more verbose, but I think it's good to try and stick with either `import` or `require` rather than have a mix of the two?

util/set-callback-urls.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const callbackUrls = async () => {
```
Gweaton, line 12: I'd like us to rely a bit more type safety in our code, do you think we could add the return type to this function (is it `Promise<string[]>`?).

util/set-callback-urls.ts
```diff
@@ -0,0 +1,55 @@
...
+    `https://api.heroku.com/pipelines/${process.env['HEROKU_PIPELINE_ID']}/review-apps`,
+    {
+      headers: {
+        Accept: 'application/vnd.heroku+json; version=3',
+        Authorization: `Bearer ${process.env['HEROKU_ACCESS_TOKEN']}`,
+      },
+    },
+  );
+
+  return response.data.map((app: any) => {
```
Gweaton, line 23: Again, re: type safety, `response.data` is of type `any` - would it be possible to declare a type for response like:
```
interface Response {
  data: string[] // or whatever this is?
}
```

Then we should be able to use `axios.get` with a Type parameter like `axios.get<Response>()` above.

Here's where I've been reading about it https://stackoverflow.com/questions/57629111/how-to-use-a-type-for-the-response-from-axios-get

Let me know if it's too much effort/not worth the trouble - it'd be nice to reduce the number of `any`s in our code if we can.

Gweaton (commented):
This looks good - I've got a few annoying TypeScript comments, let me know if I'm being too fussy! Basically, I'd like us to be able to reduce the number of `any`s in our code as it makes us less type-safe!

util/set-callback-urls.ts
```diff
@@ -0,0 +1,55 @@
+/*
+ * This is a script that gets run after a successful deploy of a Heroku
+ * review app to register the callback URLs on Auth0 to allow logins to
+ * work properly on Heroku review apps.
+ */
+const { ManagementClient } = require('auth0');
```
Gweaton, line 6: One more thing, do you think we could install the `Auth0` types library here too, so the compiler knows a bit more about the types here: `npm i --save-dev @types/auth0` (my editor highlighted this line and recommended doing this)

util/set-callback-urls.ts
pezholio: I had to do `import * as dotenv from 'dotenv'` here, as `import dotenv freom 'dotenv'` gave me an error:

```
dotenv.config({ path: '.env.development' });
       ^
TypeError: Cannot read properties of undefined (reading 'config')
```

util/set-callback-urls.ts
pezholio: Bryan had already done this in his PR, so this is in now 👍

Gweaton (approved):
Nice one, thanks for making those changes!

util/set-callback-urls.ts
pezholio: This is really cool - I like it!

PR #25, opened 2021-12-02
(1761) Add roles to users

pezholio:
This adds a new `roles` field to the user model, and adds a new proof of concept endpoint that can only be accessed by an `admin` role. These roles are not in any way set in stone, and are subject to change.


.github/workflows/ci.yml
```diff
@@ -15,6 +15,8 @@ env:
   APP_SECRET: ${{ secrets.APP_SECRET }}
   AUTH_USERNAME: ${{ secrets.AUTH_USERNAME }}
   AUTH_PASSWORD: ${{ secrets.AUTH_PASSWORD }}
+  EDITOR_USERNAME: ${{ secrets.EDITOR_USERNAME }}
```
bryangaledxw, line 18: Worth renaming `AUTH_PASSWORD` to `ADMIN_PASSWORD` for symmetry?

views/admin/superadmin.njk
```diff
@@ -0,0 +1,6 @@
+{% extends "./base.njk" %}
```
bryangaledxw, line 1: Is this the right location? Your suggestion was for views was `$MODULE/$CONTROLLER/$ACTION`

src/common/roles.decorator.ts
```diff
@@ -0,0 +1,10 @@
+import { SetMetadata } from '@nestjs/common';
+import { UserRole } from '../users/user.entity';
+
+export const Roles = (...roles: string[]) =>
```
bryangaledxw, line 4: Could this just take an array of the UserRole enum, rather than needing to map from string to enum?

src/common/unauthorized-exception.filter.ts
```diff
@@ -2,7 +2,7 @@ import {
   ExceptionFilter,
```
bryangaledxw, line 2: Need to also update JSDocs

src/common/authentication.guard.ts
```diff
@@ -11,19 +18,43 @@ import { Injectable, CanActivate, ExecutionContext } from '@nestjs/common';
...
+      request.oidc.isAuthenticated();
+
+    if (!loggedIn) {
+      throw new UnauthorizedException();
+    }
+
+    if (!roles) {
+      return loggedIn;
+    } else {
+      const user = request.appSession.user;
```
bryangaledxw, line 51: Suggest put `as User` here, then use `User` rather than `any` below

src/common/authentication.guard.spec.ts
```diff
@@ -32,6 +45,33 @@ describe('AuthenticationGuard', () => {
     it('should return true', () => {
       expect(guard.canActivate(host)).toStrictEqual(true);
     });
+
+    describe('when roles are specified', () => {
+      beforeEach(() => {
+        roles = [UserRole['ADMIN']];
```
bryangaledxw, line 51: Can all these be `User.ADMIN`/`User.EDITOR`?

src/users/user.entity.ts
```diff
@@ -1,5 +1,10 @@
 import { Entity, Column, PrimaryGeneratedColumn } from 'typeorm';
 
+export enum UserRole {
+  ADMIN = 'admin',
```
bryangaledxw, line 4: Do we want to follow an uppercase convention of enums? If so, I'll change my PR

views/admin/superadmin.njk
pezholio: This is just a temporary thing. I imagine we'll have this in a more sensible place going forward

src/users/user.entity.ts
pezholio: I cargo-culted this from somewhere else. Have changed it to Pascal-case now, which I think is better 😃

.github/workflows/ci.yml
pezholio: Good plan 👍

src/users/users.controller.spec.ts
```diff
@@ -7,6 +7,7 @@ import { createMock, DeepMocked } from '@golevelup/ts-jest';
 const name = 'Example Name';
 const email = 'name@example.com';
 const externalIdentifier = 'example-external-identifier';
+const roles = [];
```
bryangaledxw, line 10: Can this be `new Array<Role>()` so we have a stronger type than any[]?

src/users/users.seeder.ts
```diff
@@ -23,8 +23,11 @@ export class UsersSeeder implements Seeder {
     /* eslint-disable @typescript-eslint/no-var-requires */
     const userData = require('../../seeds/users.json') as SeedUser[];
 
-    const users = userData.map((user) => {
-      return new User(user.email, user.name, user.externalIdentifier);
+    const users: Array<User> = userData.map((user: any) => {
```
bryangaledxw, line 26: Think this has clashed with some of my work. Suggest get rid of the `:any` and instead add `roles: string[]` to `SeedType` above.

src/users/users.seeder.ts
```diff
@@ -23,8 +23,11 @@ export class UsersSeeder implements Seeder {
...
+      const roles = user.roles.map((role: string) => {
+        return role as UserRole;
+      });
```
bryangaledxw, line 29: Wondering if it's just as good to have this entire block as `const roles = user.roles as UserRole[]`. Either way, we're asserting that these strings match the enum

bryangaledxw (commented):
Just a few minor comments on the typing

src/users/users.controller.spec.ts
pezholio: 👍

src/users/users.seeder.ts
pezholio: Whoops - missed that!

bryangaledxw (approved):

PR #26, opened 2021-12-03
(1931) Add Qualifications model and link to Professions

Gweaton:
Adds `Qualifications` model. This represents the required qualifications necessary for a profession in the UK. We initially captured only the `qualification_level` on a `Profession`, but to pave the way for building the functionality to add a new Profession, we're adding these additional fields now to support the following page in the near future:

image

### DIFF
--- a/src/qualification/qualification.service.spec.ts
+++ b/src/qualification/qualification.service.spec.ts
@@ -1,0 +1,63 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { getRepositoryToken } from '@nestjs/typeorm';
+import { Repository } from 'typeorm';
+
+import { Qualification } from './qualification.entity';
+import { QualificationService } from './qualification.service';
+
+const qualification = new Qualification(
+  'ATT - Attestation of competence , Art. 11 a',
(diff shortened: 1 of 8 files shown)


src/qualification/qualification.service.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+import { getRepositoryToken } from '@nestjs/typeorm';
+import { Repository } from 'typeorm';
+
+import { Qualification } from './qualification.entity';
+import { QualificationService } from './qualification.service';
+
+const qualification = new Qualification(
+  'ATT - Attestation of competence , Art. 11 a',
+  ' Accredited Certification Scheme accredited by the United Kongdom Accrediation Service UKAS in accordance with ISO EN 17024 or an approprate National/Scottish Vocational Qualification at level 2',
+  ' General secondary education',
```
pezholio, line 11: Is there a reason for this leading space here?

pezholio (approved):
LGTM - just a question about the leading space in the tests. Not a huge issue though 👍

src/qualification/qualification.service.spec.ts
Gweaton: Oops, copied and pasted from the old data and missed this, will tidy up!

PR #31, opened 2021-12-07
Pluralise `Users` and `Qualifications`

bryangaledxw:
Pluralise `Users` and `Qualifications` for consistency with existing modules


Gweaton (approved):
Thanks for tidying this up!

PR #33, opened 2021-12-07
In User entity, rename `identifier` to `externalIdentifier`

bryangaledxw:
Rename for clarity of purpose of `externalIdentifier` in addition to `id`


Gweaton (approved):
Looks good - you'll need to update those failing user tests to use the newly-named field

PR #34, opened 2021-12-07
Add Professions, Qualifications, and Legislations Seeders

bryangaledxw:


src/legislations/legislations.seeder.ts
```diff
@@ -0,0 +1,32 @@
...
+import { Legislation } from './legislation.entity';
+
+@Injectable()
+export class LegislationsSeeder implements Seeder {
+  constructor(
+    @InjectRepository(Legislation)
+    private readonly legislationsRepository: Repository<Legislation>,
+  ) {}
+
+  async seed(): Promise<any> {
```
Gweaton, line 16: Do you think we could make the return type on this function `Promise<Legislation[]>` rather than `Promise<any>`? I'd like us to try and reduce the number of `any`'s in our codebase. I realise I put this in my PR too so I'm a huge hypocrite, but will update it when I get a sec!

src/professions/professions.seeder.ts
```diff
@@ -0,0 +1,72 @@
...
+    private readonly professionsRepository: Repository<Profession>,
+    @InjectRepository(Industry)
+    private readonly industriesRepository: Repository<Industry>,
+    @InjectRepository(Qualification)
+    private readonly qualificationsRepository: Repository<Qualification>,
+    @InjectRepository(Legislation)
+    private readonly legislationsRepository: Repository<Legislation>,
+  ) {}
+
+  async seed(): Promise<any> {
```
Gweaton, line 25: Again, can this be `Promise<Profession[]>`?

src/professions/professions.seeder.ts
```diff
@@ -0,0 +1,72 @@
...
+    /* eslint-disable @typescript-eslint/no-var-requires */
+    const professionsData = require('../../seeds/professions.json');
+
+    const professions: Array<Profession> = await Promise.all(
+      professionsData.map(async (profession: any) => {
```
Gweaton, line 30: I think `profession`'s type can be `Profession` here, if not we might want to define our own interface in this class to reduce our reliance on `any`.

src/industries/industries.seeder.ts
```diff
@@ -27,6 +27,6 @@ export class IndustriesSeeder implements Seeder {
   }
 
   async drop(): Promise<any> {
-    return this.industryRepository.delete({});
+    // Deletion of industries is handled in `ProfessionsSeeder`
```
Gweaton, line 30: Does Nest / TypeORM require us to even have a `drop` method if it's returning nothing? Is it possible to remove this or does it break things? If it's possible without, we could add a comment to the top of this file to explain the deletion stuff.

Gweaton (commented):
This looks good - I've just got a couple of hypocritical questions about our use of `any` that I'd like us to try and stamp out if possible, as it means we don't get the type safety that we want with TypeScript.

src/industries/industries.seeder.ts
bryangaledxw: Yep, necessary to satisfy the `Seeder` interface

src/professions/professions.seeder.ts
bryangaledxw: Needs to be `Promise<any>` to satisfy the interface. The `return` could just as easily be an `await`, as the caller isn't doing anything with the result of the Promise once it resolves

src/legislations/legislations.seeder.ts
bryangaledxw: Same as below

src/professions/professions.seeder.ts
Gweaton: argh, that's a pain.

Gweaton (approved):

PR #35, opened 2021-12-08
Professional details page

bryangaledxw:
This PR:
* Setups up an endpoint for displaying professional details
* Displays professional detail following the design [here](https://rpr-prototype.herokuapp.com/beta-sprint-2/public-facing-v1/profession-1-plastic-surgery) 
* Established canonical user-friendly URLs, and redirects if the canonical URL is not used

This PR punts on the problem of where the back link should go. Given this needs to work without Javascript, we'll need to keep track of where the user has come from. It may be that this is worthy of its own story, to handle back links across the service. If a page has multiple entry points, it may be non-trivial where 'Back' should go.

Where we don't yet have data, I've err-ed towards making it obvious we're using placeholder data, rather than risk dummy static data getting left in the page.

### DIFF
--- a/src/helpers/slug.helper.spec.ts
+++ b/src/helpers/slug.helper.spec.ts
@@ -1,0 +1,25 @@
+import { generateSlug } from './slug.helper';
+
+describe('generateSlug', () => {
+  it('should generate a URL-safe slug', () => {
+    const result = generateSlug('Example String');
+
+    expect(result).toEqual('example-string');
+  });
+
+  it('should truncate long strings to 100 characters', () => {
+    const result = generateSlug(
+      'A Very Very Very Very Very Very Very Very Very Very Very Very Very Very Very Very Very Very Very Very Long Name',
+    );
+
+    expect(result).toEqual(
+      'a-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-ver',
+    );
+  });
+
+  it('should remove non-alphanumeric characters', () => {
+    const result = generateSlug('👻 Department 😜 of 🥸 Emoji 🙀');
+
+    expect(result).toEqual('department-of-emoji');
+  });
+});
--- a/src/professions/professions.controller.spec.ts
+++ b/src/professions/professions.controller.spec.ts
@@ -7,6 +7,8 @@ import { ProfessionsService } from './professions.service';
 import { Response } from 'express';
 import { NotFoundException } from '@nestjs/common';
 
+const exampleProfession = new Profession('Example Profession');
+
 describe('ProfessionsController', () => {
   let controller: ProfessionsController;
   let professionsService: DeepMocked<ProfessionsService>;
@@ -40,13 +42,14 @@ describe('ProfessionsController', () => {
 
     it('should render a professions details page', async () => {
       professionsService.find.mockImplementationOnce(async () => {
-        return new Profession('Example Profession');
+        return exampleProfession;
       });
 
       await controller.show('example-profession', 'example-id', res);
 
       expect(res.render).toBeCalledWith('professions/show', {
-        professionName: 'Example Profession',
+        profession: exampleProfession,
+        backUrl: '',
       });
 
       expect(professionsService.find).toBeCalledWith('example-id');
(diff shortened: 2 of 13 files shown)


src/professions/professions.controller.ts
```diff
@@ -1,9 +1,23 @@
...
 
 @Controller()
 export class ProfessionsController {
+  constructor(private professionsService: ProfessionsService) {}
+
   @Get('professions/:slug/:id')
-  read(@Param('slug') slug: string, @Param('id') id: string): {slug: string, id: string} {
-    return {slug, id};
+  @Render('professions/read')
+  async read(
```
pezholio: Can we call these `show` instead of read, please

src/professions/professions.controller.ts
```diff
@@ -1,9 +1,23 @@
...
+    @Param('slug') slug: string,
+    @Param('id') id: string,
+  ): Promise<ReadTemplate> {
+
+    const profession = await this.professionsService.find(id);
+
+    if (!profession) {
+      throw new Error(`A profession with ID ${id} could not be found`);
```
pezholio, line 20: I think we can throw a `NotFoundException` here, which gives us the added bonus of being able to rescue from the 404 with a nice custom page.

src/professions/professions.controller.spec.ts
```diff
@@ -18,8 +30,28 @@ describe('ProfessionsController', () => {
...
+    });
+
+    it('should throw an error when the ID does not match a profession', () => {
+      professionsService.find.mockImplementationOnce(async () => {
+        return null;
+      });
+
+      expect(async () => {
+        await controller.read('example-slug', 'example-invalid-id');
+      }).rejects.toThrowError();
```
pezholio, line 79: We can also probably check the `NotFoundException` gets thrown if we use that instead

src/helpers/slug.helper.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+      'A Very Very Very Very Very Very Very Very Very Very Very Very Very Very Very Very Very Very Very Very Long Name',
+    );
+
+    expect(result).toEqual(
+      'a-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-ver',
+    );
+  });
+
+  it('should remove non-alphanumeric characters', () => {
+    const result = generateSlug('👻 Department 😜 of 🥸 Emoji 🙀');
```
pezholio, line 21: 😆  - love it!

src/professions/professions.controller.ts
```diff
@@ -1,23 +1,32 @@
...
     const profession = await this.professionsService.find(id);
 
     if (!profession) {
       throw new Error(`A profession with ID ${id} could not be found`);
     }
 
-    return { professionName: profession.name };
+    const generatedSlug = generateSlug(profession.name);
+
+    if (generatedSlug !== slug) {
```
pezholio, line 25: I'm a bit iffy about us having urls with IDs and slugs TBH. I'd much rather we generated a unique slug at the point of creation, added an index, and just had the slug in the URL. Makes for nicer URLs and also means we don't have to do this check.

pezholio (commented):
Looks on the right track, just a couple of comments 👍

src/professions/professions.controller.ts
bryangaledxw: My mild pushback is that if a body or profession changes its name, we have to choose between
- Broken URLs
- Slugs out of sync with names
- Logic for tracking past slugs

And my intuition is that as soon as `slug` is its own column in the DB, someone's going to want a way to edit it and we've now got am extra tricky thing to explain in our UI

pezholio: I think for now, just having one slug is enough, and having out of sync names is fine. My intuition is that names are going to change pretty irregulary, and if it begins to be a problem, we can cross that bridge when we come to it. 

Looking at a Ruby gem we use (Friendly ID), it deals with slug changes by having a history table, and searching that history table if the requested slug isn't available, then doing a 301 redirect. I think we could do something like this without too much issue, but I'd say it's out of scope for MVP.

PR #36, opened 2021-12-08
(1933) Add internationalisation

pezholio:
This moves all the text used in the views in the application to use i18n translations. This is bestpractice because:

a) It allows us to be DRY and if something changes in one place, we can change it elsewhere
b) It sets us up to be able to use other languages (e.g. Welsh) in the future
c) It makes it easier for content designers to change things in the app without digging through template files.

### DIFF
--- a/src/helpers/i18n.helper.ts
+++ b/src/helpers/i18n.helper.ts
@@ -22,9 +22,13 @@ export class I18nHelper {
    * Get a translated string from the i18n reference
    *
    * @param {text} - The translation reference (e.g `user.greeting`)
+   * @param {any} - Any personalisations that need to be passed tho the translation
    * @returns {string} - The translated string (e.g `Hello world!`)
    */
-  public async translate(text: string): Promise<string> {
-    return await this.i18nService.translate(text);
+  public async translate(
+    text: string,
+    personalisation: any = {},
+  ): Promise<string> {
+    return await this.i18nService.translate(text, { args: personalisation });
   }
 }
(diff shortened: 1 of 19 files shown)


src/helpers/i18n.helper.ts
```diff
@@ -22,9 +22,13 @@ export class I18nHelper {
    * Get a translated string from the i18n reference
    *
    * @param {text} - The translation reference (e.g `user.greeting`)
+   * @param {any} - Any personalisations that need to be passed tho the translation
    * @returns {string} - The translated string (e.g `Hello world!`)
    */
-  public async translate(text: string): Promise<string> {
-    return await this.i18nService.translate(text);
+  public async translate(
```
Gweaton, line 28: I think there's a bit of a typo in the commit message introducing this function (

> > ```
> // user.json
> {
>   "greeting": "Hi {user}"
> }
> ```
> 
> We cant then pass the user's name to the
> translation like so:
> 
> ```
> "user. greeting" | t({name: "stuart"})
> ```
> 
> This will output `Hello stuart` to the template

Should this ^ be "Hi stuart"?

Also, while you're updating, there's an extra `t` on "can[t]" in there!

Gweaton (approved):
This is great, nice one. 

Have you had any thoughts about how to get it working with our database models etc? Probably a silly question, but is there a requirement to have user-entered data e.g. professions, translated into Welsh?

src/helpers/i18n.helper.ts
pezholio: Whoops! Fat fingers there!

PR #42, opened 2021-12-13
(1925) Implement generic error pages

pezholio:
This adds a new Exception filter to catch all of the HTTP Exceptions and render an appropriate page, or redirect as is necessary. I've also removed the old exception filters that are now surplus to requirements.


src/common/http-exception.filter.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @UseFilters(HttpExceptionFilter)
+ *
+ */
+@Catch(HttpException)
+export class HttpExceptionFilter implements ExceptionFilter {
+  async catch(exception: HttpException, host: ArgumentsHost) {
+    const ctx = host.switchToHttp();
+    const response = ctx.getResponse<Response>();
+
+    if (exception.name === 'UnauthorizedException') {
```
bryangaledxw, line 26: Do we definitely want to do name string comparison rather than `exception instanceof UnauthorizedException`?

src/common/http-exception.filter.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * This class is a NestJS ExceptionFilter that catches all
+ * `HttpException`s and renders an error page.
+ *
+ * @example (in a controller, as metadata above a controller action)
+ *
+ * @UseFilters(HttpExceptionFilter)
+ *
+ */
+@Catch(HttpException)
```
bryangaledxw, line 20: If this is our one global filter, do we want to do `@Catch()`, so it catches _everything_? I tried to get something like that working, but it seemed non-trival to do something like "catch everything that hasn't already been caught", but could maybe be done with some co-ordination between the exception filters, if we can rely on the ordering.

bryangaledxw: Off the top of my head, could ValidationExceptionFilter do something like:

`exception['iHaveAlreadyBeenCaught'] = true;`

And we check for that flag in our global exception handler. (But obviously the actually clean version of this)

src/common/http-exception.filter.ts
pezholio: Ah, yeah. I tried that initially, but because I was using stubs, it wasn't working as expected in the tests. That said, I don't think there's any reason why I can't use the actual objects here. I'll give it a crack.

src/common/http-exception.filter.ts
pezholio: Yeah, we could, but I think it's tidier and easier to follow if we only catch exceptions of a particular type, to save misdirection. I think this is OK, because we're only catching a limited subsection of errors that would otherwise result in a strange user experience.

bryangaledxw (approved):
Approved, though do we still need a story to figure out how to show a nice error for the exceptions we're not currently handling? Say, things we don't know about that might come from the deepest depths of TypeORM, for instance?

pezholio:
I would expect that to bubble up and create a 500 though right?

PR #44, opened 2021-12-13
Professions details page

bryangaledxw:
This PR:
* Setups up an endpoint for displaying professional details via `professions/{professionSlug}`
* Displays professional detail following the design [here](https://rpr-prototype.herokuapp.com/beta-sprint-2/public-facing-v1/profession-1-plastic-surgery)
* Adds a `slug` column to the `Profession` entity
* Adds a method to `ProfessionsService` for adding a new profession with a unique slug, and updates existing code for creating a profession to use this

This PR punts on the problem of where the back link should go. Given this needs to work without Javascript, we'll need to keep track of where the user has come from. It may be that this is worthy of its own story, to handle back links across the service. If a page has multiple entry points, it may be non-trivial where 'Back' should go.

Where we don't yet have data, I've err-ed towards making it obvious we're using placeholder data, rather than risk dummy static data getting left in the page.

### DIFF
--- a/src/db/migrate/1639393653586-AddSlugToProfession.ts
+++ b/src/db/migrate/1639393653586-AddSlugToProfession.ts
@@ -1,0 +1,27 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddSlugToProfession1639393653586 implements MigrationInterface {
+  name = 'AddSlugToProfession1639393653586';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "professions" ADD "slug" character varying NOT NULL`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "professions" ADD CONSTRAINT "UQ_c60993e6f4badf770634fd393c7" UNIQUE ("slug")`,
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_c60993e6f4badf770634fd393c" ON "professions" ("slug") `,
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `DROP INDEX "public"."IDX_c60993e6f4badf770634fd393c"`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "professions" DROP CONSTRAINT "UQ_c60993e6f4badf770634fd393c7"`,
+    );
+    await queryRunner.query(`ALTER TABLE "professions" DROP COLUMN "slug"`);
+  }
+}
--- a/src/helpers/slug.helper.ts
+++ b/src/helpers/slug.helper.ts
@@ -10,12 +10,14 @@ const maxLength = 100;
  *   slug
  * @returns The slug generated from the provided name
  */
-export function generateSlug(name: string): string {
-  return slugify(name, {
+export function generateSlug(name: string, retryCount: number): string {
+  const base = slugify(name, {
     remove: /[^a-zA-Z0-9 ]/,
     replacement: '-',
     lower: true,
     strict: true,
     trim: true,
   }).slice(0, maxLength);
+
+  return retryCount === 0 ? base : `${base}-${retryCount}`;
 }
--- a/src/helpers/slug.helper.ts
+++ b/src/helpers/slug.helper.ts
@@ -1,0 +1,21 @@
+import slugify from 'slugify';
+
+const maxLength = 100;
+
+/**
+ * Transforms a name string into a URL-safe slug
+ *
+ * @param name The name to generate a slug from
+ * @param retryCount The number of times we've attempted to generate a unique
+ *   slug
+ * @returns The slug generated from the provided name
+ */
+export function generateSlug(name: string): string {
+  return slugify(name, {
+    remove: /[^a-zA-Z0-9 ]/,
+    replacement: '-',
+    lower: true,
+    strict: true,
+    trim: true,
+  }).slice(0, maxLength);
+}
--- a/src/i18n/en/app.json
+++ b/src/i18n/en/app.json
@@ -9,5 +9,7 @@
   "errors": {
     "heading": "There is a problem"
   },
-  "backToDashboard": "Back to the dashboard"
+  "backToDashboard": "Back to the dashboard",
+  "yes": "Yes",
+  "no": "No"
 }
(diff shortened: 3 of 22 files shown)


src/professions/professions.controller.ts
```diff
@@ -1,10 +1,33 @@
...
+
   @Get('admin/professions/add-profession')
   @Render('professions/admin/add-profession/new')
   new(): Record<string, any> {
     return {};
   }
+
+  @Get('professions/:id')
+  @Render('professions/show')
+  async show(@Param('id') id: string): Promise<ShowTemplate> {
```
pezholio: Do we need to import `ShowTemplate` here, or does Typescript pick this up automagically?

src/professions/professions.controller.spec.ts
```diff
@@ -38,14 +40,15 @@ describe('ProfessionsController', () => {
...
+      professionsService.find.mockImplementationOnce(
+        async () => exampleProfession,
+      );
 
       const result = await controller.show('example-id');
 
       expect(result).toEqual({
-        professionName: 'Example Profession',
+        profession: exampleProfession,
+        backUrl: '',
```
pezholio, line 51: Can we get the `backUrl` from `req.get('Referrer')`? We could also put a check in to see if it's come from our domain or not and optionally not show the back link if someone has got here from elsewhere (i.e. a deep link)

src/db/migrate/1639393653586-AddSlugToProfession.ts
```diff
@@ -0,0 +1,27 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddSlugToProfession1639393653586 implements MigrationInterface {
+  name = 'AddSlugToProfession1639393653586';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "professions" ADD "slug" character varying NOT NULL`,
```
pezholio, line 8: This `NOT NULL` causes an issue when the database already has data, which is annoying. We can get around this for now, as the data we have isn't super important, but we need to be careful about adding `NOT NULL` migrations when we have production data, and the database has data that we actually care about

src/professions/professions.service.ts
```diff
@@ -19,11 +21,47 @@ export class ProfessionsService {
...
 
-  async create(profession: Profession): Promise<void> {
-    await this.repository.insert(profession);
-  }
-
   findBySlug(slug: string): Promise<Profession> {
     return this.repository.findOne({ where: { slug } });
   }
+
+  async replaceSlugAndAdd(profession: Profession): Promise<Profession> {
```
pezholio, line 28: I still think I'd rather call this `create`, as it better encapsulates what's being done. `replaceSlugAndAdd` describes the implementation details, not the overall intent.

src/i18n/en/app.json
```diff
@@ -9,5 +9,9 @@
   "errors": {
     "heading": "There is a problem"
   },
-  "backToDashboard": "Back to the dashboard"
+  "backToDashboard": "Back to the dashboard",
+  "yesOrNo": {
```
pezholio, line 13: I think we can just have `yes` and `no` a level up here. I don't think it's necessary to embed it in `yesOrNo`

pezholio (commented):
Looks good so far, just a couple of comments 👍

src/db/migrate/1639393653586-AddSlugToProfession.ts
bryangaledxw: ✅

src/professions/professions.controller.ts
bryangaledxw: Yep that should have had an import

Gweaton (commented):
I think this looks good - have just added a couple of small comments. I agree with Stu's comments re: naming of `replaceSlugAndAdd`, so I'll leave this as a "comment" rather than "approve".

views/professions/show.njk
```diff
@@ -9,14 +11,214 @@
...
+        
+        {{ govukSummaryList({
+          classes: 'govuk-summary-list--no-border',
+          rows: [
+            {
+              key: {
+                text: "Reserved activities"
+              },
+              value: {
+                html: ["<ul class=\"govuk-list\"><li>", (escapedActivites | join("</li><li>")), "</li></ul>"] | join
```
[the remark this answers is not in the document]
bryangaledxw: I have no idea if this is the idiomatic way of doing this 😅

src/helpers/slug.helper.ts
```diff
@@ -6,14 +6,22 @@ const maxLength = 100;
...
+export function generateSlug(name: string, retryCount: number): string {
+  const base = slugify(name, {
     remove: /[^a-zA-Z0-9 ]/,
     replacement: '-',
     lower: true,
     strict: true,
     trim: true,
   }).slice(0, maxLength);
+
+  if (retryCount == 0) {
```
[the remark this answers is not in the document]
bryangaledxw: Yep, should be. Thoughts on ternaries are that I like them when they improve readability, which I get is very very subjective. For me this is a 50/50 case, but will go with a ternary

src/professions/professions.service.spec.ts
```diff
@@ -43,8 +44,17 @@ const professionArray = [
...
   let service: ProfessionsService;
   let repo: Repository<Profession>;
+  let manager: DeepMocked<EntityManager>;
 
   beforeEach(async () => {
+    const queryRunner = createMock<QueryRunner>();
+    const connection = createMock<Connection>();
+
+    manager = createMock<EntityManager>();
+    Object.assign(queryRunner, { manager });
```
Gweaton, line 54: Is there a way to mock `queryRunner.manager` rather than assigning the whole object here? Using `Object.assign` in a test feels like something we can get Jest to mock for us. I've had a bit of a playaround, but haven't been able to get it working yet @pezholio I don't suppose you have any ideas?

src/professions/professions.service.spec.ts
```diff
@@ -107,4 +112,92 @@ describe('Profession', () => {
       expect(repoSpy).toHaveBeenCalledWith({ where: { slug: 'some-slug' } });
     });
   });
+
+  describe('create', () => {
+    it('should save the profession with the "base" slug when there are no colliding slugs', async () => {
+      const profession = new Profession();
```
Gweaton, line 118: Can we pass `name` here rather than setting it on the line below (and in the test below)?

Gweaton (commented):
A couple of small things, dunno if @pezholio might be able to think of anything for one of them?

src/professions/professions.service.spec.ts
bryangaledxw: This definitely wasn't my first choice of approach - the problem here is that manager is `readonly`. I've pushed a small improvement that dodges the need for `Object.assign`. It does seem that `DeepMocked` is something of a misnomer here. I'm wondering if it's because `manager` is a property, which I've run into problems with with mocking before

Gweaton (approved):
Looks better with that mocked version of `manager` thanks for figuring it out!

PR #46, opened 2021-12-14
Fix saving professions

Gweaton:
Switch to using `.save`, not `.insert` to create Professions

`.insert` only creates the specified entity, and doesn't run any
callbacks or create any relations, so the Industry ID wasn't being set
as a foreign key here.

Switching to `.save` exposed a flaw in our understanding too, as we
can't create a new `Qualification` here (or any other foreign keys) as
we do in the `Profession` entity's constructor. We've updated the
constructor to set foreign keys as `null` instead when a value isn't
supplied. We'll soon be able to add the real `Qualification`,
`Legislation` etc, so won't need to create a new instance of an object
here anyway - we'll be looking it up.

### DIFF
--- a/src/professions/admin/add-profession/confirmation.controller.ts
+++ b/src/professions/admin/add-profession/confirmation.controller.ts
@@ -7,7 +7,6 @@ import {
   Session,
 } from '@nestjs/common';
 import { IndustriesService } from '../../../industries/industries.service';
-import { Qualification } from '../../../qualifications/qualification.entity';
 import { Profession } from '../../profession.entity';
 
 import { ProfessionsService } from '../../professions.service';
@@ -36,7 +35,7 @@ export class ConfirmationController {
       topLevelDetails.nation,
       '',
       industry,
-      new Qualification(),
+      null,
       [],
       [],
     );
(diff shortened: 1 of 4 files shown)


src/professions/professions.service.ts
```diff
@@ -19,7 +19,7 @@ export class ProfessionsService {
     return this.repository.findOne(id);
   }
 
-  async create(profession: Profession): Promise<void> {
-    await this.repository.insert(profession);
+  async create(profession: Profession): Promise<Profession> {
+    return await this.repository.save(profession);
```
bryangaledxw, line 23: I think `return await` is redundant, can just do `return`

src/professions/admin/add-profession/confirmation.controller.ts
```diff
@@ -36,12 +35,12 @@ export class ConfirmationController {
       topLevelDetails.nation,
       '',
       industry,
-      new Qualification(),
-      [],
-      [],
+      null,
+      null,
+      null,
```
bryangaledxw, line 40: I'd still keep these last two as empty lists. Let's use valid data where we can, even if it is only temporary

bryangaledxw (approved):
Approved with comments

PR #47, opened 2021-12-14
Support internationalisation of industry names

bryangaledxw:
This PR allows internationalisation of industry names. I've kept the approach as consistent Nations as possible

### DIFF
--- a/seeds/industries.json
+++ b/seeds/industries.json
@@ -1,26 +1,26 @@
 [
   {
-    "name": "Construction & Engineering"
+    "name": "industries.constructionAndEngineering"
   },
   {
-    "name": "Education"
+    "name": "industries.education"
   },
   {
-    "name": "Finance"
+    "name": "industries.finance"
   },
   {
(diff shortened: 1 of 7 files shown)


Gweaton (approved):

seeds/industries.json
```diff
@@ -1,26 +1,26 @@
 [
   {
-    "name": "Construction & Engineering"
+    "name": "industries.constructionAndEngineering"
```
pezholio, line 3: Do we need `industries` here? It might make more sense to have this specified in the template, so we're more explicit about what's going on.

bryangaledxw: If I'm getting rid of industries here, then I should get rid of it for nations, then I'm messing around with a bunch of existing templates and the `tOptionSelect` filter, and it felt like the wrong rabbit hole to be going down when search is a priority

pezholio: OK, cool. Let's leave it as is then 👍

PR #48, opened 2021-12-15
Use database rather than session when creating a user

pezholio:
As discussed, we're moving on from storing details of a multi-page form in a session to in the database. This does mean we might run the risk of having partially created objects in the database, but this is a risk worth taking. We can also balance this by having a clean up script running regularly (if it becomes a problem).


bryangaledxw (commented):
I've not fully reviews, but giving this a try there's a lot that worries me

- I don't know how we're meant to filter out partially created users when we're doing anything that only wants to consider fully created users. Should there be a column in the DB so we can mark users as "partial"?

- As a practical example of the above, if someone doesn't complete creating a user, and comes back later, they now can't create a user with that email address, as the "no duplicate emails" check looks at both complete and partial users

- Do we need some logic for cleaning out partially created users if they've not been fully created within some time period? So the db doesn't get polluted if someone doesn't complete the creation flow

- If someone knows the database ID of a user, they can use this flow to edit an existing user

- Should the back button from the confirm screen be working?

- We went with more logic to avoid having a DB ID in the URL for the professions page, so I feel iffy about us now having a DB ID in the URL here

- I understand that "new" makes sense from a consistency point of view, but I don't feel like it's the nicer URL from the user's point of view. Looking a the PCR/Lateral test flows again, those go with long, descriptive URLs

- I don't know that we can get away without the "edit" query param when we have more pages and want to enforce that the user goes through the flow in the correct order _but_ allow them to jump to any point once they've reached the confirm screen

pezholio:
> I don't know how we're meant to filter out partially created users when we're doing anything that only wants to consider fully created users. Should there be a column in the DB so we can mark users as "partial"?

Probably - I think that's a "cross that bridge when we come to it" issue

> As a practical example of the above, if someone doesn't complete creating a user, and comes back later, they now can't create a user with that email address, as the "no duplicate emails" check looks at both complete and partial users

This can we something we handle as part of the edit journey I think

> If someone knows the database ID of a user, they can use this flow to edit an existing user

That's fine - we want that to be able to happen eventually. We just need to make sure we only allow people with the right permissions to do this.

> Should the back button from the confirm screen be working?

It should be. At least it worked last time I checked 😬 

> We went with more logic to avoid having a DB ID in the URL for the professions page, so I feel iffy about us now having a DB ID in the URL here

Back office URLs I'm less concerned about having IDs in, especially when they're create/edit type URLs. When a URL has information in thats likely to be linked to around the web, it's more important that they're stable.

> I understand that "new" makes sense from a consistency point of view, but I don't feel like it's the nicer URL from the user's point of view. Looking a the PCR/Lateral test flows again, those go with long, descriptive URLs

Can you give me an example here?

> I don't know that we can get away without the "edit" query param when we have more pages and want to enforce that the user goes through the flow in the correct order but allow them to jump to any point once they've reached the confirm screen

I think there's a limited risk here, because we're dealing with backend users that we trust. We should probably validate the whole user object at the end of the journey though.

bryangaledxw:
So as an example URL - `https://test-for-coronavirus.service.gov.uk/order-lateral-flow-kits/condition` - where it "action you are doing"/"flow step". Though yeah, it's not as lengthy as I was remembering

My high-level worry is that not having a clear divide between partial and complete users sets up more bridges for us to cross in future than we had before. And I think a lot of the simplification is just coming from getting rid of the logic that was anticipating a longer multistep flow and keeping tight control of the user journey through that flow

src/users/user.entity.ts
```diff
@@ -16,7 +16,8 @@ export class User {
   @Column()
   name: string;
 
-  @Column({ unique: true })
+  @Index({ unique: true, where: '"externalIdentifier" IS NOT NULL' })
```
bryangaledxw, line 19: I'm surprised `unique` isn't implicit in `Index()` - does a non-unique index make sense?

src/users/personal-details/personal-details.controller.ts
```diff
@@ -0,0 +1,87 @@
...
+import { backLink } from '../../common/utils';
+
+@Controller('/admin/users')
+export class PersonalDetailsController {
+  constructor(private readonly usersService: UsersService) {}
+
+  @Get(':id/personal-details/edit')
+  @UseGuards(AuthenticationGuard)
+  @Render('users/personal-details/edit')
+  async edit(@Req() req: Request, @Param('id') id): Promise<any> {
```
bryangaledxw, line 29: I've been trying to setup interfaces that correspond to what a particular template expects. Worth setting one up for this?

src/users/personal-details/personal-details.controller.ts
```diff
@@ -0,0 +1,87 @@
...
+    };
+  }
+
+  @Post(':id/personal-details')
+  @UseGuards(AuthenticationGuard)
+  async create(
+    @Body() personalDetailsDto,
+    @Res() res,
+    @Param('id') id,
+  ): Promise<object> {
```
bryangaledxw, line 44: Does `Promise<void>` work here?

src/users/users.controller.ts
```diff
@@ -24,63 +19,48 @@ export class UsersController {
...
-  @Post('/admin/users/create-new-user/confirm')
+  @Post('/admin/users/:id/confirm')
   @UseGuards(AuthenticationGuard)
-  async create(@Session() session, @Res() res): Promise<object> {
-    const userCreationFlowSession = new UserCreationFlowSession(
-      session,
-      UserCreationFlowStep.AllDetailsEntered,
-    );
-    const { email, name } = userCreationFlowSession.sessionDto;
+  async complete(@Res() res, @Param('id') id): Promise<object> {
```
bryangaledxw, line 48: Similar comment as above about setting up an interface to correspond to the template

src/common/utils.ts
```diff
@@ -0,0 +1,10 @@
+import { Request } from 'express';
+
+export const backLink = (req: Request): string => {
```
bryangaledxw, line 3: Suggest do as `export function backLink()`

src/common/utils.ts
```diff
@@ -0,0 +1,10 @@
...
+  const referrer = req.get('Referrer') || '';
+  const host = req.get('host');
+
+  if (referrer.includes(host)) {
```
bryangaledxw, line 7: Would `startsWith` be safer and more explicit about what we're checking for? (I acknowledge this is being very paranoid)

src/users/personal-details/personal-details.controller.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+import { createMock, DeepMocked } from '@golevelup/ts-jest';
+
+const name = 'Example Name';
+const email = 'name@example.com';
+
+describe('PersonalDetailsController', () => {
+  let controller: PersonalDetailsController;
+  let usersService: DeepMocked<UsersService>;
+  let user: User;
+  let populatedSession;
```
bryangaledxw, line 16: This can be deleted now

src/users/personal-details/personal-details.controller.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+    );
+  });
+
+  it('should be defined', () => {
+    expect(controller).toBeDefined();
+  });
+
+  describe('edit', () => {
+    it('should get and return the user from an id', async () => {
+      const request = createMock<Request>({
```
bryangaledxw, line 57: Worth pulling out? We're going to want something like this for every page that has a back link, I think - something like `function createMockRequest(host: string)` or even just hardcode it to `example.com`

bryangaledxw (commented):
Few minor comments, and I acknowledge the one about setting up interfaces for templates is something inherited from my code (I've since been following George's example on that), and I don't regard those as blockers

Have checked and the back button isn't working for me on the confirm page - it leaves off the `edit` from the URL. Not sure what going on, wondering if the redirect is confusing things, happy to approve once that's working

src/users/personal-details/personal-details.controller.ts
pezholio: Oooh, that's a good idea. It's a new one on me, but I think it sounds like something that's worth doing 👍

src/users/users.controller.ts
pezholio: I'm a bit confused by this one - what interface should we have? Because we're returning different things depending on the context (and sometimes not returning at all)

bryangaledxw: Ah, this can just be `Promise<void>` I think

src/users/personal-details/personal-details.controller.spec.ts
pezholio: 👍

src/users/users.controller.ts
```diff
@@ -24,63 +20,51 @@ export class UsersController {
...
-  async create(@Session() session, @Res() res): Promise<object> {
-    const userCreationFlowSession = new UserCreationFlowSession(
-      session,
-      UserCreationFlowStep.AllDetailsEntered,
-    );
-    const { email, name } = userCreationFlowSession.sessionDto;
+  async complete(
+    @Res() res,
+    @Param('id') id,
+  ): Promise<String | UserAlreadyExists> {
```
bryangaledxw, line 52: I feel like I'm missing something? This function never actually returns anything, just returns early or finishes, so Promise<void> should be correct? Also the linter doesn't like capital-S `string`

bryangaledxw (approved):
Cool, back buttons works now. There's an unresolved comment from first review about a `Promise<object>` that can be a promise void, and a couple of new ones. Also, able to fix the linter warnings before merging?

Also, the case I flagged early is worse than I thought - if you go to the confirm screen, then go back just to edit the name but not the email, you can't do that because you instantly get a validation error. If that's a non-issue for now, happy for this to go in

src/users/personal-details/interfaces/user-with-back-link.ts
```diff
@@ -0,0 +1,5 @@
+import { User } from '../../user.entity';
+
+export interface UserWithBackLink extends User {
```
bryangaledxw, line 3: Sorry - type this out then I think it go lost. Is this the naming we want to go with? I've tried to keep the naming to {TemplateName}Template - to make it obvious what the purpose is, and make it clear what needs to change if this changes. Though I recognise there's some cases that'll be super generic and re-used

src/users/users.controller.ts
pezholio: This may well be me getting the wrong end of the stick - I didn't know whether the `req.render` would be classed as a `return` by the linter. `void` is better 😃

bryangaledxw: Cool. Was surprised the type system is fine either way 🤷‍♀️

pezholio: Same!

PR #49, opened 2021-12-15
(1951) Select multiple nations and industries when adding a profession

Gweaton:
Allows users to select multiple nations & industries when adding a profession via checkboxes, replacing the original Option Select.

Updates the wording a bit on some of the screens for more detail.

## Screenshots

image


src/professions/admin/add-profession/check-your-answers.controller.ts
```diff
@@ -23,11 +27,16 @@ export class CheckYourAnswersController {
       topLevelDetails.industryId,
     );
 
-    const selectedNation = Nation.find(topLevelDetails.nation);
+    const selectedNations: string[] = await Promise.all(
+      topLevelDetails.nations.map(async (nationCode) => {
+        const nationKey = Nation.find(nationCode).name;
 
+        return await this.i18nService.translate(nationKey);
```
pezholio, line 40: Could we add this to the `Nation` class, so we have a method that returns the translated name (i.e `nation.translatedName` or something)

seeds/professions.json
```diff
@@ -5,7 +5,7 @@
     "description": "Registration protection and exploitation of trade marks",
```
pezholio, line 5: There's a typo in the commit message: `Profesion` vs `Profession`

src/professions/profession.entity.ts
```diff
@@ -59,7 +60,7 @@ export class Profession {
     this.description = description || '';
     this.occupationLocations = occupationLocations || [];
     this.regulationType = regulationType || '';
-    this.industry = industry || null;
+    this.industries = industries || null;
```
pezholio, line 63: Should this default to an empty array?

cypress/integration/admin/professions/add-a-profession.spec.ts
```diff
@@ -13,9 +13,8 @@ describe('Adding a new profession', () => {
     );
 
     cy.get('input[name="name"]').type('Example Profession');
-    // TODO: figure out how to select a single checkbox, as this is currently throwing CORS issues
-    cy.get('[type="checkbox"]').check();
-    cy.get('select[name="industryId"]').select('Construction & Engineering');
+    cy.get('[type="checkbox"]').check('GB-ENG');
+    cy.contains('Construction & Engineering').click();
```
pezholio, line 17: Can we use the translation to look this up?

src/professions/admin/add-profession/check-your-answers.controller.spec.ts
```diff
@@ -35,22 +35,29 @@ describe('CheckYourAnswersController', () => {
...
       };
 
-      const industry = new Industry('Construction & Engineering');
+      const industry = new Industry('industries.construction');
       industry.id = constructionUUID;
 
-      industriesService.find.mockImplementation(async () => industry);
+      industriesService.findByIds.mockImplementation(async () => [industry]);
+      // This is a bit ugly, but there's no way to mock with arguments as far as I'm aware
+      // so we'll need to rely on ordering of mocks here.
```
pezholio, line 48: You can do something like this with the service:

```
i18nService = createMock<I18nService>({
      translate: async (text) => {
        if (text === 'industries.construction') {
          return 'Construction & Engineering';
        }
      },
    });
```

pezholio (approved):
LGTM - just a couple of little things 👍

bryangaledxw:
Flagging that this'll need a change to the professional details page, where we now have multiple industries - shout if you need an hand

bryangaledxw:
Also, was this flagged with/coming from Ming? I think a lot of the designs right now are assuming a single industry

src/professions/admin/add-profession/check-your-answers.controller.ts
Gweaton: That would be good actually, do you think it'd be weird to pass a `translationService` to the `Nation` in the constructor?

src/professions/profession.entity.ts
Gweaton: This makes TypeORM unhappy, I think it's something to do with it being a foreign key, rather than an array of strings, for example, like we have above. I get:

```
InitializedRelationError: Array initializations are not allowed in entity relations. Please remove array initialization (= []) from "Profession#industries". This is ORM requirement to make relations to work properly. Refer docs for more information.
```

Will leave it as `null` for now.

Gweaton:
> Also, was this flagged with/coming from Ming? I think a lot of the designs right now are assuming a single industry

Have run this past Ming and got the thumbs up!

> Flagging that this'll need a change to the professional details page, where we now have multiple industries - shout if you need a hand

This was tricky, but I think I've rebased correctly.

src/professions/admin/add-profession/check-your-answers.controller.ts
Gweaton: Have passed into the function call instead, but not 100% sure about it, what do you think?

pezholio: Aaah, I see what you mean. I get why you're not 100%, but I think it's clear enough 👍

src/professions/profession.entity.ts
pezholio: 👍

PR #51, opened 2021-12-16
(1929) Allow roles to be specified against user in UI

pezholio:
This adds a Roles screen to the Create User journey. Now we have more than one screen to contend with, this has complicated the change journey a bit, so I've made some tweaks to make sure users get redirected to the correct screen after making changes.

## Screenshots

image

image


src/validation/validation-exception.filter.ts
```diff
@@ -28,10 +28,12 @@ import { ValidationFailedError } from './validation-failed.error';
 export class ValidationExceptionFilter implements ExceptionFilter {
   view: string;
   objectName: string;
+  additionalVariables: any;
 
-  constructor(view: string, objectName: string) {
+  constructor(view: string, objectName: string, additionalVariables?: any) {
```
Gweaton, line 33: Nice one, I don't think this will work with more involved variables, as I couldn't figure out how pass things like entities we lookup from the db to a decorator, but thanks for getting this in. Can we do any better than `any` here? `Record<string, unknown>` might be an option.

views/users/confirm.njk
```diff
@@ -37,6 +37,11 @@
 
         <h2 class="govuk-heading-m">{{ "users.form.headings.personalDetails" | t }}</h2>
 
+          {% set roleItems = "" %}
+          {% for role in roles %}
+            {% set translation = "users.form.label." + role %}
+            {% set roleItems = roleItems + (translation | t) + "<br/>" %}
```
Gweaton, line 43: Does this work? I found translations in loops in templates threw some nunjucks filter errors, which is why I ended up translating things in the controller / classes. I wonder if it works because you're appending to the `roleItems` here instead? Is this something you're planning to move into the `presenter` / `view` classes in future?

Gweaton (approved):
Looks good - a couple of minor things.

views/users/confirm.njk
pezholio: It works, but yeah, watch this space for further improvements!

src/validation/validation-exception.filter.ts
pezholio: Cool - done!

PR #52, opened 2021-12-16
Remove test code from the utils module

rjw1:
`createMockRequest` relies on a dev dependency which fails production builds.
To fix this we have split it out to its own module and told `nest build` to
ignore it in `tsconfig.build.json`.


pezholio (approved):
Ah, sorry, that was my fault 😞. LGTM though 👍

PR #53, opened 2021-12-16
Add Timestamps to all database entities

pezholio:
This is best practice, and allows us to better keep track of when data has been created and/or updated.


Gweaton (approved):
Good shout - you'll probably want to rebase this now that I've merged my updates to the Profession entity

PR #56, opened 2021-12-17
Add empty `docker-entrypoint`

Gweaton:
We could choose to run seeds in here, or some other setup stuff if we
need to. It's the standard way of writing Dockerfiles.


pezholio (approved):
LGTM 👍

PR #57, opened 2021-12-17
Unfiltered professions search page (1812)

bryangaledxw:
This PR adds an unfiltered professions search page, following Ming's prototype - https://rpr-prototype.herokuapp.com/beta-sprint-2/public-facing-v1/selected-service

### DIFF
--- a/cypress/integration/professions/search/search.spec.ts
+++ b/cypress/integration/professions/search/search.spec.ts
@@ -1,0 +1,45 @@
+describe('Searching a profession', () => {
+  it('I can view an unfiltered list of profession', () => {
+    cy.visit('/professions/search');
+
+    cy.get('body').should('contain', 'Registered Trademark Attorney');
+    cy.get('body').should(
+      'contain',
+      'Secondary School Teacher in State maintained schools (England)',
+    );
+  });
+
+  it('I can click a profession to be taken to its details page', () => {
+    cy.visit('/professions/search');
+
+    cy.get('a')
+      .contains(
+        'Secondary School Teacher in State maintained schools (England)',
+      )
+      .click();
+    cy.url().should(
+      'contain',
+      'professions/secondary-school-teacher-in-state-maintained-schools-england',
+    );
+  });
+
+  it('I can enter search filters', () => {
+    cy.visit('/professions/search');
+
+    cy.get('input[name="keywords"]').type('Attorney');
+    cy.get('input[name="nations"][value="GB-ENG"]').check();
+    cy.get('input[name="nations"][value="GB-NIR"]').check();
+
+    cy.get('input[name="industries"]').eq(0).check();
+    cy.get('input[name="industries"]').eq(2).check();
+
+    cy.get('button').click();
+
+    cy.get('input[name="keywords"]').should('have.value', 'Attorney');
+    cy.get('input[name="nations"][value="GB-ENG"]').should('be.checked');
+    cy.get('input[name="nations"][value="GB-NIR"]').should('be.checked');
+
+    cy.get('input[name="industries"]').eq(0).should('be.checked');
+    cy.get('input[name="industries"]').eq(2).should('be.checked');
+  });
+});
--- a/src/professions/search/search.controller.ts
+++ b/src/professions/search/search.controller.ts
@@ -1,0 +1,106 @@
+import { Body, Controller, Get, Post, Render, Req } from '@nestjs/common';
+import { Request } from 'express';
+import { I18nService } from 'nestjs-i18n';
+import { backLink } from '../../common/utils';
+import { IndustriesService } from '../../industries/industries.service';
+import { Nation } from '../../nations/nation';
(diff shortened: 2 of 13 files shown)


cypress/integration/professions/search/search.spec.ts
```diff
@@ -0,0 +1,31 @@
+describe('Searching a profession', () => {
+  it('I can view an unfiltered list of profession', () => {
+    cy.visit('/professions/search');
+
+    cy.get('body').should('contain', 'Registered Trademark Attorney');
```
pezholio, line 5: Can we use the translations here?

src/professions/search/search.controller.ts
```diff
@@ -0,0 +1,106 @@
...
+    private readonly professionsService: ProfessionsService,
+    private readonly industriesService: IndustriesService,
+    private readonly i18nService: I18nService,
+  ) {}
+
+  @Get()
+  @Render('professions/search/search')
+  async search(@Req() request: Request): Promise<SearchTemplate> {
+    return {
+      ...(await this.creatSearchResults(new FilterDto())),
```
pezholio, line 23: Typo 👍

src/professions/search/search.controller.ts
```diff
@@ -0,0 +1,106 @@
...
+    @Body() filter: FilterDto,
+    @Req() request: Request,
+  ): Promise<SearchTemplate> {
+    return {
+      ...(await this.creatSearchResults(filter)),
+      backLink: backLink(request),
+    };
+  }
+
+  private async creatSearchResults(
```
pezholio, line 40: Could we break this out into a `SearchService`? I think I'd prefer the various different variables that get returned at the end to be broken out into their own methods too, to make it more easier to test and to follow. It'll then make the controller tests simpler and seperate out the concerns a bit.

pezholio: That said, I know you're working on the next bit now, so I'm happy for this to go in as is and we can iterate 👍

views/professions/search/search.njk
```diff
@@ -0,0 +1,182 @@
...
+      
+      <p class="govuk-body-m"><span class="govuk-!-font-size-36 govuk-!-font-weight-bold">{{ professions.length }}</span> {{ "profession" if (professions.count === 1) else "professions" }} found.</p>
+
+      <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
+
+      <div class="rpr-listing__profession-container">
+
+        {% for profession in professions %}
+          <h2>
+            <a class="govuk-link govuk-heading-l rpr-listing__profession-title" href="profession-1-plastic-surgery">{{profession.name}}</a>
```
pezholio, line 116:
```suggestion
<a class="govuk-link govuk-heading-l rpr-listing__profession-title" href="/professions/{{ profession.id }}">{{profession.name}}</a>
```

src/professions/search/search.controller.ts
```diff
@@ -0,0 +1,106 @@
...
+@Controller('professions/search')
+export class SearchController {
+  constructor(
+    private readonly professionsService: ProfessionsService,
+    private readonly industriesService: IndustriesService,
+    private readonly i18nService: I18nService,
+  ) {}
+
+  @Get()
+  @Render('professions/search/search')
```
pezholio, line 20: I think `index` is a better name for this template

src/professions/search/search.controller.ts
```diff
@@ -0,0 +1,106 @@
...
+  async search(@Req() request: Request): Promise<SearchTemplate> {
```
pezholio, line 21: Again, as this is the root action `index` is the better name here

src/professions/search/search.controller.ts
```diff
@@ -0,0 +1,106 @@
...
+    return {
+      ...(await this.creatSearchResults(new FilterDto())),
+      backLink: backLink(request),
+    };
+  }
+
+  @Post()
+  @Render('professions/search/search')
+  async searchPost(
```
pezholio, line 30: This can probably be called `create`, as we're `create`ing a search query.

pezholio (approved):
Just a couple of typos and naming considerations. One thing to think about, but I know you're working on the next thing, so happy for this to go in as is and iterate on the next card 👍

cypress/integration/professions/search/search.spec.ts
bryangaledxw: I don't think so? - this is the name in the DB and the idea of localised names for professions isn't really something we've tackled

pezholio: Oh, yeah, you're right. Carry onj 😄

src/professions/search/search.controller.ts
bryangaledxw: Yep, this is the first think in the upcoming PR

views/professions/search/search.njk
bryangaledxw: Woops!

PR #66, opened 2021-12-21
(1963) Persist draft Professions in the database, rather than the session

Gweaton:
We now create a blank profession at the start of the journey, fetch it on each page via the ID in the URL params (which we're happy with using, rather than a slug, as this is internal-facing behaviour), and update it.

This will make it much easier for us to:
- Build functionality for us to edit Professions in the future
- Add changelinks

It also paves the way with a pattern for adding the remaining screens to the Add a profession journey.

Still to come, in future PRs:
- Change links on the "Check your answers" page
- Functional back buttons
- Additional pages
- Put this all behind authentication guards

Apologies for the size of one of the commits here, I hope it's not too painful to review.

### DIFF
--- a/src/professions/admin/add-profession/check-your-answers.controller.ts
+++ b/src/professions/admin/add-profession/check-your-answers.controller.ts
@@ -3,6 +3,7 @@ import { I18nService } from 'nestjs-i18n';
 
 import { Nation } from '../../../nations/nation';
 import { ProfessionsService } from '../../professions.service';
+import { CheckYourAnswersTemplate } from './interfaces/check-your-answers.template';
 
 @Controller('admin/professions')
 export class CheckYourAnswersController {
@@ -13,12 +14,7 @@ export class CheckYourAnswersController {
 
   @Get(':id/check-your-answers')
   @Render('professions/admin/add-profession/check-your-answers')
-  async show(@Param('id') id: string): Promise<{
-    name: string;
-    nations: string[];
-    industries: string[];
-    professionId: string;
-  }> {
+  async show(@Param('id') id: string): Promise<CheckYourAnswersTemplate> {
     const draftProfession = await this.professionsService.find(id);
 
     if (!draftProfession) {
--- a/src/professions/admin/add-profession/check-your-answers.controller.ts
+++ b/src/professions/admin/add-profession/check-your-answers.controller.ts
@@ -1,26 +1,25 @@
-import { Controller, Get, Render, Session } from '@nestjs/common';
+import { Controller, Get, Param, Render } from '@nestjs/common';
 import { I18nService } from 'nestjs-i18n';
 
 import { Nation } from '../../../nations/nation';
 import { ProfessionsService } from '../../professions.service';
 
-@Controller('admin/professions/new/check-your-answers')
+@Controller('admin/professions')
 export class CheckYourAnswersController {
   constructor(
     private readonly professionsService: ProfessionsService,
     private readonly i18nService: I18nService,
   ) {}
 
-  @Get()
+  @Get(':id/check-your-answers')
   @Render('professions/admin/add-profession/check-your-answers')
-  async show(@Session() session: Record<string, any>): Promise<{
+  async show(@Param('id') id: string): Promise<{
     name: string;
     nations: string[];
     industries: string[];
+    professionId: string;
   }> {
-    const professionId = session['profession-id'];
-
-    const draftProfession = await this.professionsService.find(professionId);
+    const draftProfession = await this.professionsService.find(id);
 
     if (!draftProfession) {
       throw new Error('Draft profession not found');
@@ -39,6 +38,7 @@ export class CheckYourAnswersController {
     );
 
     return {
+      professionId: id,
       name: draftProfession.name,
       nations: selectedNations,
       industries: industryNames,
--- a/src/professions/admin/add-profession/check-your-answers.controller.ts
+++ b/src/professions/admin/add-profession/check-your-answers.controller.ts
@@ -1,14 +1,13 @@
 import { Controller, Get, Render, Session } from '@nestjs/common';
 import { I18nService } from 'nestjs-i18n';
 
-import { IndustriesService } from '../../../industries/industries.service';
 import { Nation } from '../../../nations/nation';
-import { TopLevelDetailsDto } from './dto/top-level-details.dto';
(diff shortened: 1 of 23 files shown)


src/professions/admin/add-profession/top-level-information.controller.ts
```diff
@@ -1,44 +1,78 @@
...
+import { ProfessionsService } from '../../professions.service';
 import { TopLevelDetailsDto } from './dto/top-level-details.dto';
+import { TopLevelDetailsTemplate } from './interfaces/top-level-details.template';
 
-@Controller('admin/professions/new/top-level-information')
+@Controller('admin/professions')
 export class TopLevelInformationController {
-  constructor(private industriesService: IndustriesService) {}
+  constructor(
+    private professionsService: ProfessionsService,
```
bryangaledxw, line 23: I've probably missed this plenty of times, but I'm trying to use `readonly` here

src/professions/admin/add-profession/top-level-information.controller.ts
```diff
@@ -1,44 +1,78 @@
...
+
     const industries = await this.industriesService.all();
 
-    const industriesCheckboxArgs = industries.map((industry) => ({
-      text: industry.name,
-      value: industry.id,
-    }));
+    const industriesCheckboxArgs = industries.map((industry) => {
+      const hasSelectedIndustry = profession.industries
+        ? profession.industries.find(
```
bryangaledxw, line 39: I think `foo !== undefined` can be `!!foo`

src/professions/admin/add-profession/interfaces/top-level-details.template.ts
```diff
@@ -0,0 +1,6 @@
+export interface TopLevelDetailsTemplate {
+  name: string | null;
+  industriesCheckboxArgs: { text: string; value: string; checked?: boolean }[];
```
bryangaledxw, line 3: Do we gain anything from `checked` being optional? And can this just use `CheckboxArgs`? Ditto below

src/professions/admin/add-profession/check-your-answers.controller.ts
```diff
@@ -1,46 +1,41 @@
...
-import { TopLevelDetailsDto } from './dto/top-level-details.dto';
+import { ProfessionsService } from '../../professions.service';
+import { CheckYourAnswersTemplate } from './interfaces/check-your-answers.template';
 
-@Controller('admin/professions/new/check-your-answers')
+@Controller('admin/professions')
 export class CheckYourAnswersController {
   constructor(
-    private industriesService: IndustriesService,
+    private professionsService: ProfessionsService,
```
bryangaledxw, line 11: Similar comment to before out `readonly`

src/professions/admin/add-profession/confirmation.controller.ts
```diff
@@ -1,56 +1,31 @@
...
+  @Get('/:id/confirmation')
   @Render('professions/admin/add-profession/confirmation')
-  async viewConfirmation(@Session() session: Record<string, any>) {
-    const addProfessionSession = session['add-profession'];
+  async viewConfirmation(
+    @Param('id') id: string,
+  ): Promise<ConfirmationTemplate> {
+    const profession = await this.professionsService.find(id);
+
+    this.professionsService.confirm(profession);
```
bryangaledxw, line 27: I think I'm confused about the double confirmation - that it's both here and in the `@Post` method. Also, calls to `confirm` should be `await`ed. This isn't the kind of operation we want happening on a `@Get`, right?

src/professions/admin/add-profession/confirmation.controller.ts
```diff
@@ -1,56 +1,31 @@
...
-      [],
-      [],
-    );
-
-    await this.professionsService.create(profession);
+  constructor(private professionsService: ProfessionsService) {}
+
+  @Post('/:id/confirmation')
+  async confirm(@Res() res: Response, @Param('id') id: string): Promise<void> {
+    const profession = await this.professionsService.find(id);
```
bryangaledxw, line 13: Worth a check here to make sure we're not trying to confirm an already confirmed profession? (Or maybe have that inside `confirm`?

src/professions/admin/add-profession/confirmation.controller.ts
```diff
@@ -1,56 +1,28 @@
...
-    private industriesService: IndustriesService,
-  ) {}
+  constructor(private professionsService: ProfessionsService) {}
 
-  @Post()
-  @Redirect('/admin/professions/new/confirmation')
-  async confirm(@Session() session: Record<string, any>) {
-    const draftProfession = await this.fetchProfessionOrThrow(session);
+  @Post('/:id/confirmation')
+  async confirm(@Res() res: Response, @Param('id') id: string) {
```
pezholio: Could this be called `create`?

src/professions/admin/add-profession/confirmation.controller.ts
```diff
@@ -1,56 +1,28 @@
...
+
+    res.redirect(`/admin/professions/${id}/confirmation`);
   }
 
-  @Get()
+  @Get('/:id/confirmation')
   @Render('professions/admin/add-profession/confirmation')
-  async viewConfirmation(@Session() session: Record<string, any>) {
-    const profession = await this.fetchProfessionOrThrow(session);
+  async viewConfirmation(@Param('id') id: string) {
```
pezholio: Could this be called `new`?

src/professions/admin/add-profession/top-level-information.controller.ts
bryangaledxw: And less efficient but tidier I think - does `(profession.industries || []).find()` work?

pezholio (approved):
Lovely stuff. Just a couple of naming things, and don't forget the Changelog 😸

bryangaledxw (commented):
Looks good. Blockers for approval for me are my comments on the confirmation controller.

Also, with the back link, we've got the `backLink` utility function now, though now worries if you still want to leave it. Also, I think you've got an unclosed tag on the confirmation page - visible if you a look at the footer

src/professions/admin/add-profession/confirmation.controller.ts
Gweaton: Yeah, this was a mistake - well done for spotting!

bryangaledxw (approved):

PR #71, opened 2021-12-22
(1972) Send email to new users

pezholio:
This adds [Bull](https://github.com/OptimalBits/bull) to allow us to queue and process jobs, as well as using Bull to send emails via GOV.UK Notify.

I've taken the approach we have had in other projects and added a very simple template to GOV.UK Notify, with `(( subject ))` and `(( body ))` variables in like so:

image

This allows us to use the same template across the project, as well as keeping our email body content tightly coupled to the code and version controlled.


src/users/auth0-user-creation.service.ts
```diff
@@ -51,6 +45,24 @@ export class Auth0UserCreationService extends ExternalUserCreationService {
...
+      result: 'user-created',
+      externalIdentifier: user.user_id,
+      passwordResetLink: passwordChangeTicket.ticket,
+    };
+  }
+
+  private getClient(domain: string) {
+    return new ManagementClient({
+      domain,
+      clientId: process.env.AUTH0_CLIENT_ID,
```
bryangaledxw, line 64: Suggest change these to `process.env['...']`, so we're not mixing and matching in the same file

bryangaledxw (approved):
Looks good, approved with minor comment

bryangaledxw:
Able to add `dump.rdb` to `.gitignore` as part of this?

src/users/auth0-user-creation.service.ts
pezholio: D'oh. I'm really bad at remembering to do this!

PR #72, opened 2021-12-22
Fix invalid profession submit

Gweaton:
# Changes in this PR

There was an issue here where the page would hang on submitting without
a required field, due to there not being a `@Render` decorator on the
`edit` function. I've extracted the rendering functionality into its own
function, and tweaked it the logic for what to display.

The tests are a bit complex, and there might be a better way to handle
this, but here's what the new logic does:
- If there's a blank Profession (i.e. we've just landed on the page)
  persist user input even if one or more fields are missing, so they
  don't have to re-type their answers.
- If there's already a filled-in Profession, e.g. the user is editing
  after already completing this step, render those answers on the page,
  but if a new answer is submitted and a required field is missing,
  render the new values in the form, rather than the older ones on the
  Profession.

### DIFF
--- a/src/professions/admin/add-profession/top-level-information.controller.spec.ts
+++ b/src/professions/admin/add-profession/top-level-information.controller.spec.ts
@@ -83,46 +83,49 @@ describe('TopLevelInformationController', () => {
       it('should fetch all Industries and Nations to be displayed in an option select, with none of them checked', async () => {
         professionsService.find.mockImplementation(async () => blankProfession);
 
-        const templateArgs = await controller.edit('profession-id');
+        await controller.edit(response, 'profession-id');
 
-        expect(templateArgs).toEqual({
-          name: null,
-          industriesCheckboxArgs: [
-            {
-              text: 'Health',
-              value: 'health-uuid',
-              checked: false,
-            },
-            {
-              text: 'Construction & Engineering',
-              value: 'construction-uuid',
-              checked: false,
-            },
-          ],
-          nationsCheckboxArgs: [
-            {
-              text: 'England',
-              value: 'GB-ENG',
-              checked: false,
-            },
-            {
-              text: 'Scotland',
-              value: 'GB-SCT',
-              checked: false,
-            },
-            {
-              text: 'Wales',
-              value: 'GB-WLS',
-              checked: false,
-            },
-            {
-              text: 'Northern Ireland',
-              value: 'GB-NIR',
-              checked: false,
-            },
-          ],
-          errors: undefined,
-        });
+        expect(response.render).toHaveBeenCalledWith(
+          'professions/admin/add-profession/top-level-information',
+          {
+            name: null,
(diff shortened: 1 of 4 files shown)


src/professions/admin/add-profession/top-level-information.controller.ts
```diff
@@ -29,35 +21,20 @@ export class TopLevelInformationController {
...
     errors: object | undefined = undefined,
-  ): Promise<TopLevelDetailsTemplate> {
+  ): Promise<void> {
     const profession = await this.professionsService.find(id);
 
-    const industries = await this.industriesService.all();
-
-    const industriesCheckboxArgs = await new IndustriesCheckboxPresenter(
-      industries,
+    return this.renderTopLevelInformation(
```
pezholio, line 31: Given this is in the `TopLevelInformationController`, we can probably call this `renderForm`

src/professions/admin/add-profession/top-level-information.controller.ts
```diff
@@ -96,6 +84,68 @@ export class TopLevelInformationController {
...
+    profession: Profession,
+    topLevelDetailsDto: TopLevelDetailsDto,
+  ): Promise<Industry[]> {
+    if (topLevelDetailsDto.industries) {
+      return this.industriesService.findByIds(topLevelDetailsDto.industries);
+    }
+
+    if (profession.industries) {
+      return profession.industries;
+    }
```
pezholio, line 132: Could we not just do a ternery here? Something like:

```
return topLevelDetailsDto.industries ? this.industriesService.findByIds(topLevelDetailsDto.industries) : profession.industries;
```

Might just tighten it up a bit

src/professions/admin/add-profession/top-level-information.controller.spec.ts
```diff
@@ -218,23 +224,489 @@ describe('TopLevelInformationController', () => {
...
     });
 
-    describe('when required parameters are not entered', () => {
-      it('does not create a profession, and re-renders the top level information view with errors', async () => {
-        const topLevelDetailsDtoWithNoAnswers = {
-          name: '',
-          nations: undefined,
-          industries: undefined,
-        };
+    describe('when required fields are not entered', () => {
```
pezholio, line 227: Wonder whether we can simplify these tests by unit testing the individual helper methods, rather than doing what is essentially a functional test for every single edge case. We'd have to make those methods public, but I don't think it's a big deal.

pezholio (commented):
I think the overall approach is sound, but I wonder if we can simplify the tests by dropping down a level on the unit tests for our edge cases.

src/professions/admin/add-profession/top-level-information.controller.spec.ts
Gweaton: I think I've simplified them a bit!

pezholio (approved):
Yay - this is great 👍

PR #76, opened 2021-12-23
(1950) Add page for selecting a regulatory body

Gweaton:
# Changes in this PR

- Adds a screen for selecting a regulatory body when adding a new profession.
- Adds seeds for Organisations, currently not linked to any professions.
- Adds regulatory body data to check your answers page.

#### Add regulatory / professional bodies

image

#### Check your answers

image

### DIFF
--- a/seeds/organisations.json
+++ b/seeds/organisations.json
@@ -1,0 +1,42 @@
+[
+  {
+    "name": "Council of Registered Gas Installers",
+    "alternateName": "",
+    "address": "123 Fake Street, London, EC1 1AB",
+    "url": "www.corgi-gas-safety.com",
+    "email": "gas@example.com",
+    "contactUrl": "gas@example.com/contact-us",
+    "telephone": "+44 0123 456789",
+    "fax": "+44 0123 456780"
+  },
+  {
+    "name": "Department for Education",
+    "alternateName": "",
+    "address": "123 Example Street, London, EC1 1AB",
+    "url": "https://www.gov.uk/guidance/early-years-qualifications-finder#overseas-qualifications",
+    "email": "dfe@example.com",
+    "contactUrl": "def@example.com/contact-us",
+    "telephone": "+44 0123 456789",
+    "fax": "+44 0123 456780"
+  },
+  {
+    "name": "Law Society of England and Wales",
+    "alternateName": "",
+    "address": "456 Example Street, London, EC1 1AB",
+    "url": "www.lawsociety.org.uk",
+    "email": "law@example.com",
+    "contactUrl": "law@example.com/contact-us",
+    "telephone": "+44 0123 456789",
+    "fax": "+44 0123 456780"
+  },
+  {
+    "name": "General Medical Council",
+    "alternateName": "",
+    "address": "456 Fake Street, London, EC1 1AB",
+    "url": "www.gmc.com",
+    "email": "gmc@example.com",
+    "contactUrl": "gmc@example.com/contact-us",
+    "telephone": "+44 0123 456789",
+    "fax": "+44 0123 456780"
+  }
+]
--- a/seeds/professions.json
+++ b/seeds/professions.json
@@ -15,6 +15,7 @@
       "The administration of oaths."
     ],
     "legislations": ["The Trade Marks Act 1994", "Legal Services Act 2007"],
+    "mandatoryRegistration": "voluntary",
     "confirmed": true
   },
   {
@@ -31,6 +32,7 @@
       "The Education (School Teachers' Qualifications) (England) Regulations 2003/1662 (as amended)",
       "The Education (Induction Arrangements for School Teachers) (England) Regulations 2012/1115 (as amended)"
     ],
+    "mandatoryRegistration": "mandatory",
     "confirmed": true
   },
   {
--- a/src/professions/admin/add-profession/dto/regulatory-body.dto.ts
+++ b/src/professions/admin/add-profession/dto/regulatory-body.dto.ts
@@ -1,0 +1,10 @@
+import { IsNotEmpty } from 'class-validator';
+import { MandatoryRegistration } from '../../../profession.entity';
+
+export class RegulatoryBodyDto {
+  @IsNotEmpty()
+  regulatoryBody: string;
+
+  @IsNotEmpty()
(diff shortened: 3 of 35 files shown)


src/professions/admin/add-profession/regulatory-body.controller.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+  });
+
+  describe('edit', () => {
+    describe('when editing a just-created, blank Profession', () => {
+      it('should render a list of Organisations to be displayed in a Select, alongside Radio Buttons with mandatory registration values with none of them selected', async () => {
+        professionsService.find.mockImplementation(async () => profession);
+
+        await controller.edit(response, 'profession-id');
+
+        expect(response.render).toHaveBeenCalledWith(
```
bryangaledxw, line 83: Since the presenters have their own tests, I think it's ok to use them in constructing our expected results, reducing the size of these tests. I think this is what Stu did when he added User(s)Presenter

src/professions/admin/add-profession/regulatory-body.controller.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+
+    i18nService.translate.mockImplementation(async (text) => {
+      switch (text) {
+        case 'professions.form.radioButtons.mandatoryRegistration.mandatory':
+          return 'Mandatory';
+        case 'professions.form.radioButtons.mandatoryRegistration.voluntary':
+          return 'Voluntary';
+        case 'professions.form.radioButtons.mandatoryRegistration.unknown':
+          return 'Unknown';
+        default:
```
bryangaledxw, line 53: I want to suggest that the default case should just return `text`, and when mocking `I18nService`, we only mock translations that are tested *directly* in this test, rather than tested indirectly via lower-level presenters. Otherwise these are gonna grow to silly sizes with silly amounts of repetition

src/professions/profession.entity.ts
```diff
@@ -38,6 +44,9 @@ export class Profession {
   @Column({ nullable: true })
   regulationType: string;
 
+  @Column({ nullable: true, type: 'enum', enum: MandatoryRegistration })
+  mandatoryRegistration: string;
```
bryangaledxw, line 48: Can this not be `: MandatoryRegistration`?

src/professions/professions.service.ts
```diff
@@ -29,8 +29,8 @@ export class ProfessionsService {
     return this.repository.findOne({ where: { slug } });
   }
 
-  async save(user: Profession): Promise<Profession> {
```
bryangaledxw, line 32: I'm trying to stick to a convention where if I'm fixing unrelated earlier mistakes, I put those commits in first, so it's tidying, then actual work

src/professions/admin/add-profession/dto/regulatory-body.dto.ts
```diff
@@ -0,0 +1,9 @@
+import { IsNotEmpty } from 'class-validator';
+
+export class RegulatoryBodyDto {
+  @IsNotEmpty()
+  regulatoryBody: string;
+
+  @IsNotEmpty()
+  mandatoryRegistration: string;
```
bryangaledxw, line 8: Can this be the enum?

src/professions/admin/add-profession/regulatory-body.controller.ts
```diff
@@ -87,13 +90,19 @@ export class RegulatoryBodyController {
 
     await this.professionsService.save(updated);
 
+    if (regulatoryBodyDto.change) {
+      // This will be a different page once the next page in the journey is added
```
bryangaledxw, line 94: This comment should be on the final `return` down, not this one, right?

bryangaledxw (commented):
A few minor comments, the big one for me is that I feel easy about using the localisation IDs as something other than localisation IDs, would far far rather it be "mandatory", "voluntary" or "unknown" going in the DB.

Also wondering if a bunch of the places that are `string` can be the enum. Or if that doesn't work, would something like `type MandatoryRegistration = 'mandatory' | 'voluntary' | 'unknown' work? Since from a JS perspective, it will really just be a string, just with some TS side restrictions on which strings

seeds/professions.json
```diff
@@ -31,6 +32,7 @@
       "The Education (School Teachers' Qualifications) (England) Regulations 2003/1662 (as amended)",
       "The Education (Induction Arrangements for School Teachers) (England) Regulations 2012/1115 (as amended)"
     ],
+    "mandatoryRegistration": "professions.mandatoryRegistration.mandatory",
```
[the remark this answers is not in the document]
Gweaton: Yeah, I initially stored values in the DB like this, then seconds before opening the PR had a sudden thought that it might need to be translated. I think you're right, it does feel a bit weird to store values like this in the DB. I'll put it back to how it was and possibly add a helper method to fetch the right value from the localisation ID.

src/professions/admin/add-profession/regulatory-body.controller.spec.ts
Gweaton: Good point, have fixed this.

src/professions/admin/add-profession/regulatory-body.controller.spec.ts
Gweaton: Ah, I didn't spot this. Have updated to use the presenters.

src/professions/profession.entity.ts
Gweaton: Yeah I think I read the docs for this wrong, and though this was how you defined the type. Have updated this

seeds/professions.json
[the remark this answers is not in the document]
Gweaton: Ok, I've fixed this now, I hope it's not too hacky.

bryangaledxw (approved):

seeds/organisations.json
```diff
@@ -0,0 +1,42 @@
+[
```
pezholio, line 1: This will need to see in `seeds/test` and `seeds/development` now

pezholio: Oh wait, ignore me! I'll fix this on my PR 😃

pezholio (approved):

PR #78, opened 2022-01-04
(1982) Seed users on deploy

pezholio:
We currently don't have a way to seed users at deploy time, meaning we don't have any users to log into the system and start populating data. I've got around this by having environment-specific seed files, and attempted to DRY things up with a decorator that fetches the data, or returns an empty array. With this in place, we can update the Dockerfile to run the seed task before booting the app.


src/decorators/seeds.decorator.ts
```diff
@@ -0,0 +1,14 @@
+export function InjectData(filename: string) {
+  return function (target: any) {
```
Gweaton, line 2: can target be `Seeder` here as all the things that use it implement `Seeder`, or is that not what's happening here?

Gweaton (approved):
Just one comment about the usual `any` here.

Do you think we could get my branch in before this one? There's a new Seeder in there for Organisations, so it'd be good to put that in here if that's not too much of a headache.

src/decorators/seeds.decorator.ts
pezholio: I get this error if I do that 😕 

```
Property 'prototype' does not exist on type 'Seeder'.ts(2339)
```

Gweaton: Have had a play around, tried creating a new interface that extended Seeder and added a `prototype` to satisfy this, but it didn't seem to accept our own Seeder classes, even though they also implement Seeder.

Having read the TypeScript docs for decorators it looks like they use `any` here: https://www.typescriptlang.org/docs/handbook/decorators.html#decorator-composition so I think I'll just have to get over my fear of `any`s in the codebase until we have time to clear them out as much as possible.

pezholio: I've gone for another approach using a property decorator, which seems cleaner. Lemme know what you think!

Gweaton (approved):

PR #81, opened 2022-01-05
Make regulated activities strings

Gweaton:
# Changes in this PR

Updates Regulated Activities field on Professions to represent free text, rather than an array of strings, as we'd previously designed it. This matches the latest designs for adding a Profession, which includes a textarea for adding this information. In the long term, we may want to iterate on this to provide more structured data that will be easier to display to users, but for now this will work with the dataset we want to import.


pezholio (approved):
LGTM 👍

PR #84, opened 2022-01-05
Add Database seed documentation

pezholio:
Add some brief documentation about how to run and create seed data


bryangaledxw (approved):
Looks good. My only quibble is in some cases you might not want to add your new seeder to the _bottom_ of `src/seeder.ts`, if there's dependencies (as with the professions seeder). If there's a succinct way of capturing that I think it's worth mentioning, but I can't think of a way of explaining it that's less complex than just looking at the code, so not super fussed

PR #86, opened 2022-01-05
(1806) Internal professions view

bryangaledxw:
# Changes in this PR

This PR add an internal professions listing, and allows filtering by keyword, nation, organisation, and industry. The displayed columns and filters depends on the type of currently logged in user.

This also adds the following that may be useful in future:
* `stringifyNations`, which joins a list of nations, and collapses them to "United Kingdom" where applicable
* `OrganisationsCheckboxPresenter`, similar to checkbox presenter classes for nations and industries
* `createMockI18nService`, which removes some boilerplate from tests
* Professions are now seeded with an organisation
* `formatDate` to enforce a standard date format for our app

Future work will be needed to implement filtering by user, though this is blocked on users having an organisation, and to bring this in line with a to-be-agreed users permission model. We also currently display a placeholder "friendly ID"

### Before

N/A

### After

Professions listing (editor user)
localhost_3000_admin_professions_list (2)

Professions listing (admin user)
localhost_3000_admin_professions_list (1)


src/professions/admin/list/list.controller.ts
```diff
@@ -0,0 +1,121 @@
+import {
```
pezholio, line 1: I'd expect this to be `admin/professions.controller.ts`

src/professions/admin/list/list.controller.ts
```diff
@@ -0,0 +1,121 @@
...
+  @UseGuards(AuthenticationGuard)
+  @Render('professions/admin/list/index')
+  async index(@Req() request: Request): Promise<IndexTemplate> {
+    return this.createListEntries(new FilterDto(), request);
+  }
+
+  @Post()
+  @UseGuards(AuthenticationGuard)
+  @Render('professions/admin/list/index')
+  async create(
```
pezholio, line 45: I think `search` is a more logical name here, as we're not `CREATE`ing anything really.

pezholio (approved):
LGTM - just my usual naming nitpicking really 😃 

I would also have maybe split 76eba053a09aaa319edbdaf4eb76db2cd2267f2d into two parts, one for the list and one for the filtering just to make it slightly easier to follow.

I'm also a bit iffy about commits that are just for adding tests (I know sometimes it's unavoidable), because really the tests should be committed alongside the code, but don't stress it this time round 👍

Gweaton (approved):
This looks like a lot of work, well done on getting it out. Looks really good to me, just have a couple of minor comments.

Feel free to ignore the one about using Fishery in tests for now, unless you feel like taking care of it in this PR, apologies for creating more work for you!

PR #89, opened 2022-01-06
Fix Redis config

pezholio:
When we set up Redis, I assumed GOV.UK PaaS would handle it in much the same way as Postgres (i.e. add a "magic"  `REDIS_HOST` variable). 

However, there's a bit more legwork involved. When we bind a service to an app, the connection details are `VCAP_SERVICES` environment variable, which we then need to parse at runtime. This adds a new config object that parses the `VCAP_SERVICES` env var and gets the relevant connection details. This is then injected into the Bull config at runtime.

I've tweaked them to both use the redis `url` config variables for consistency, and updated the example env file, so folks may need to update their local env files to get Bull working again.


Gweaton (approved):
This makes sense to me, let's see if it works!

PR #90, opened 2022-01-06
Upload screenshots and videos when a test fails

pezholio:
This uploads the videos and screenshots of failing Cypress tests if they fail in CI. We can then download them and review them from the Github Actions Artifacts section i.e:

image

I've had this branch hanging round for ages, and I've just had to use it again, so thought it'd be useful.


Gweaton (approved):
Didn't know about this, this is cool!

PR #91, opened 2022-01-06
Add seeds for staging

pezholio:
This adds some extra seeds we can use in staging, as well as adding some scripts to make running the seed command (and npm commands more generally) easier.


Gweaton (approved):
Nice

PR #92, opened 2022-01-06
(1992) Use Fishery for creating test objects

Gweaton:
# Changes in this PR

Adds "Fishery" package which gives us factories for creating objects to use in tests, rather than the combination of normal objects and mocks that we've had up until now. This takes a lot of boilerplate out of test setup and provides a consistent approach to testing.

I've updated the majority of the tests in the codebase, using factories for:
- Organisations
- Professions
- Users
- Legislations
- Industries

The one area I've left untouched is the Search functionality tests, as I know they're currently being worked on and I don't want to cause issues with merge conflicts. We can incorporate these new factory objects into those tests as a separate piece of work once both branches are merged in.

Apologies for the sheer number of commits, most of them are very small!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -64,6 +64,7 @@
         "eslint": "^8.0.1",
         "eslint-config-prettier": "^8.3.0",
         "eslint-plugin-prettier": "^4.0.0",
+        "fishery": "^2.1.0",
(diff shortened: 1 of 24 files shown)


pezholio (approved):
Love it! Thanks for doing this 👍

PR #94, opened 2022-01-06
Get Redis options from URL

pezholio:
It seems that we can't pass the URL in directly, as that leads to it being ignored (see https://github.com/OptimalBits/bull/issues/1338), so we need to parse the individual parts from the url and pass them in as an `IORedis.RedisOptions` object.


Gweaton (approved):
Oof, well-discovered.

PR #96, opened 2022-01-07
Improve errors in Add a profession journey

Gweaton:
# Changes in this PR

- Fixes missing error highlighting on form fields when there's a validation error.
- Adds custom user-friendly error messages, rather than the standard `x should not be empty` default that comes from `class-validator`

**Note**:This is currently only being done for Professions at the moment, as there's an issue with translating errors when we have multiple error messages we want to show for a single field (e.g. trying to create a user with an invalid and missing email address) that I've spent a lot of time on today but would like to park until we have some more time to iron this out. It may be worth us listing multiple errors separately, rather than joining them with commas, as this feels a bit off to me.

### Before

image

### After

image

### DIFF
--- a/src/professions/admin/add-profession/dto/regulated-activities.dto.ts
+++ b/src/professions/admin/add-profession/dto/regulated-activities.dto.ts
@@ -1,10 +1,12 @@
 import { IsNotEmpty } from 'class-validator';
 
 export class RegulatedActivitiesDto {
-  @IsNotEmpty()
+  @IsNotEmpty({ message: 'professions.form.errors.reservedActivities.empty' })
   activities: string;
 
-  @IsNotEmpty()
+  @IsNotEmpty({
+    message: 'professions.form.errors.description.empty',
+  })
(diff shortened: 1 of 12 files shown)


src/professions/admin/add-profession/dto/regulated-activities.dto.ts
```diff
@@ -1,10 +1,12 @@
 import { IsNotEmpty } from 'class-validator';
 
 export class RegulatedActivitiesDto {
-  @IsNotEmpty()
+  @IsNotEmpty({ message: 'Enter reserved activities' })
```
pezholio, line 4: Can we use translations for these?

Gweaton: Ah yes, oops. Nunjucks makes this a bit complex due to the way we pass errors into the template so will figure out the best way to do this.

pezholio (approved):
LGTM 👍

PR #97, opened 2022-01-07
Delete user in Auth0 when deleted

pezholio:
I noticed that when we delete a user in the app, it doesn't delete them in Auth0, which can result in (scary music) **unintended consequences**. This adds a method to delete a user as a background task, as well as making a few tweaks to how we stub out behaviour with the external API.


doc/background-jobs-and-emails.md
```diff
@@ -55,3 +52,60 @@ which takes anything put in the `mailer` queue and makes an API call to GOV.UK N
...
+You can then add a method that responds with a `PerformNowOrLater` type like so:
+
+```typescript
+public myLongRunningTask(ARGS): PerformNowOrLater {
+  return {
+    performNow() => {
+      // This is what actually carries out the job
+    },
+    performLater() => {
+      return await this.queue.add('myLongRunningTask', ARGS);
```
Gweaton, line 79: Just for my understanding, does this then run the `performNow` once it's the next thing on the queue?

Gweaton (approved):
I think this makes sense, thanks for updating the documentation, it really helped with reviewing!

src/users/users.controller.spec.ts
```diff
@@ -37,7 +37,7 @@ describe('UsersController', () => {
 
     request = createMock<Request>();
 
-    userService = createMock<ExternalAuthProviderService>({
+    auth0Service = createMock<Auth0Service>({
```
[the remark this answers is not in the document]
pezholio: I'd say it's still a unit test, because we're still mocking things that aren't part of the unit, but I've had to go one level deeper, because I have to check that a) the `auth0Service` receives `deleteUser`, and b) the `deleteUser` response recieves `performLater`

doc/background-jobs-and-emails.md
pezholio: Yep 👍

PR #99, opened 2022-01-10
Don't fetch old profession values on invalid form submit

Gweaton:
# Changes in this PR

Originally, if a user went back to change their answers then tried to
submit an empty value e.g. reserved activities, we'd play back the
value on the profession in the field but highlight the field with an
error. We've decided this probably isn't the most logical user
experience, so we're going to play back an empty field.

Users will be able to see their original answer on the "Check your
answers" page by clicking the "back" button, and this shouldn't be a
common scenario, so we don't think it's a big problem.

This has a handy side effect of making tests and implementation code a
lot simpler.


bryangaledxw (approved):
Looks good!

PR #103, opened 2022-01-10
Update data types for "method to obtain" and "duration" of qualifications

Gweaton:
# Changes in this PR

Adds some database changes to pave the way for the new "Qualification Information" page, now that we know a bit more about how Qualifications should be structured:

- Uses enum fields for storing `methodToObtain` and `commonPathToObtain` values to give us consistency in responses. Adds a text `other` field for each of these, when a user needs to enter their own value.
- Adds additional, more granular `educationDurationX` fields, as the original string value wasn't really fit for purpose. We'll remove the old field when we start displaying these new fields in the service.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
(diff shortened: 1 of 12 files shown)


pezholio (approved):
Lovely stuff, I think this is the best way to play it 👍

PR #105, opened 2022-01-11
Add a service owner boolean to a user

pezholio:
This adds a new `serviceOwner` boolean to a user, allowing us to specify if a user can see content for all organisations, or just their organisation.

## Screnshots of UI changes

image

### DIFF
--- a/src/professions/admin/professions.controller.ts
+++ b/src/professions/admin/professions.controller.ts
@@ -13,7 +13,7 @@ import {
   ProfessionsPresenterView,
 } from './professions.presenter';
 import { AuthenticationGuard } from '../../common/authentication.guard';
-import { User, UserRole } from '../../users/user.entity';
+import { User } from '../../users/user.entity';
 import { FilterDto } from './dto/filter.dto';
 import { OrganisationsService } from '../../organisations/organisations.service';
 import { Organisation } from '../../organisations/organisation.entity';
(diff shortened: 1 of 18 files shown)


views/users/roles/edit.njk
```diff
@@ -24,6 +25,31 @@
...
+            isPageHeading: false,
+            classes: "govuk-fieldset__legend--l"
+          }
+        },
+        hint: {
+          text: ("users.form.hint.serviceOwner" | t)
+        },
+        items: [
+          {
+            value: 1,
```
Gweaton, line 43: I've got something similar in my branch (added a `yesAndNoRadioButtonsPresenter`) that uses "yes" and "no" for its values - what's the motivation behind 1 and 0 here? I can see the logic, I just don't think I've done it this way before.

src/professions/admin/professions.controller.ts
```diff
@@ -49,7 +49,7 @@ export class ProfessionsController {
 
     const user = request['appSession'].user as User;
 
-    const overview = user.roles.includes(UserRole.Admin);
+    const overview = user.serviceOwner;
```
Gweaton, line 52: Do you think we could we make this variable name a bit more boolean-sounding like `isOverview` or `showAllOrganisations`?

Gweaton (approved):
Looks good, just a couple of comments.

views/users/roles/edit.njk
pezholio: I guess it just seemed like a more logical way (to me anyway) to cast to a boolean. Again, proabably a carry-over from Rails.

src/professions/admin/professions.controller.ts
pezholio: Yeah, nice idea 👍

PR #106, opened 2022-01-11
Make `methodToObtain` enum values nullable to fix staging deployment

Gweaton:
# Changes in this PR
Update migration to make `methodToObtain` enum values nullable

There's an issue with staging deployments where the migrations fail due
to this field being null in the database at the time of migration. We
believe this is an issue tied to using an `enum` value for a field in
the DB (which is why we've never had issues creating new non-nullable
fields up until now) - we always need to make these nullable or provide
a default value. In this case, a default value didn't seem sensible.

We'll catch `null` values at code level, rather than database level, so
this shouldn't be a huge issue for us.

I'm editing the existing migration here, which is something we shouldn't
normally do, but if we don't, the migrations won't able to get past this
one successfully.

### DIFF
--- a/seeds/test/qualifications.json
+++ b/seeds/test/qualifications.json
@@ -18,7 +18,7 @@
     "otherMethodToObtain": "",
     "commonPathToObtain": "generalOrVocationalPostSecondaryEducation",
     "otherCommonPathToObtain": "",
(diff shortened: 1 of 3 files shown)


pezholio (approved):
LGTM 👍

PR #110, opened 2022-01-11
(1762) Define user permissions

pezholio:
This updates the `roles` attribute on a user to `permissions`, and defines a whole bunch of new, more granular permissions
to allow us to specify exactly what a user can do. I've also applied these permissions in the controller using the `Permissions` controller decorator, and also added the ability to read data about a user and their permissions in the view, so we can conditionally show/hide buttons and navigation.

### DIFF
--- a/src/i18n/en/users.json
+++ b/src/i18n/en/users.json
@@ -3,9 +3,17 @@
     "label": {
       "name": "Name",
       "email": "Email address",
-      "roles": "What roles apply to this user?",
-      "admin": "Administrator",
-      "editor": "Editor",
+      "permissions": "What permissions apply to this user?",
+      "createUser": "Create User",
+      "editUser": "Edit User",
+      "deleteUser": "Delete User",
+      "createOrganisation": "Create organisation",
+      "deleteOrganisation": "Delete organisation",
+      "createProfession": "Create profession",
+      "deleteprofession": "Delete profession",
+      "uploadDecisionData": "Upload decision data",
+      "downloadDecisionData": "Download decision data",
+      "viewDecisionData": "View decision data",
       "serviceOwner": "Is this user a service owner?"
     },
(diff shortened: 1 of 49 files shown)


src/i18n/en/users.json
```diff
@@ -3,9 +3,17 @@
...
       "name": "Name",
       "email": "Email address",
-      "roles": "What roles apply to this user?",
-      "admin": "Administrator",
-      "editor": "Editor",
+      "permissions": "What permissions apply to this user?",
+      "create-user": "Create User",
+      "edit-user": "Edit User",
+      "delete-user": "Delete User",
+      "create-organisation": "Create organisation",
```
Gweaton, line 10: Do you think we could make these camel case as we've done in the rest of the translations, rather than kebab case?

src/users/personal-details/personal-details.controller.ts
```diff
@@ -41,6 +44,7 @@ export class PersonalDetailsController {
   }
 
   @Post(':id/personal-details')
+  @Permissions(UserPermission.CreateUser, UserPermission.EditUser)
   @UseGuards(AuthenticationGuard)
```
Gweaton, line 48: Can we take this off here now that the guard is on the controller?

Gweaton (approved):
I think this approach is really good. Have just added a couple of small comments.

Also, there's a typo in the commit message `Conitionally hide create profession button ` if you feel like tidying that up!

src/users/personal-details/personal-details.controller.ts
pezholio: Oh yeah, whoops - missed that!

src/i18n/en/users.json
pezholio: Yeah, that's nicer actually 👍

PR #113, opened 2022-01-12
(1995) Public-facing regulatory authority listing

bryangaledxw:
# Changes in this PR

Adds a public-facing regulatory authority listing page, with the same filters as the public-facing professions page. The new page is at `/regulatory-authorities/search`.

This PR includes some work to avoid duplication between professions and organisations filtering. There is potentially further work that can be done, that would encompass public facing professions and organisations search, and internal professions and organisations search, but if this is something we want, I think it would be best to tackle once both this PR, and the internal organisations listing is in.

### Before

N/A

### After

localhost_3000_regulatory-authorities_search


src/organisations/organisations.service.ts
```diff
@@ -14,6 +14,10 @@ export class OrganisationsService {
     return this.repository.find();
   }
 
+  allWithProfessions(): Promise<Organisation[]> {
+    return this.repository.find({ relations: ['professions'] });
+  }
```
pezholio, line 19: I've done this in my PR by adding an options param to the `all` method, which I think is probably clearer? I'm open to being told otherwise though 😃

src/organisations/helpers/organisations-sorter.ts
```diff
@@ -0,0 +1,17 @@
+import { Organisation } from '../organisation.entity';
+
+export class OrganisationsSorter {
+  constructor(private readonly organisations: Organisation[]) {}
+
+  sortByName(): Organisation[] {
```
pezholio, line 6: This might be easier to do at the database level by passing a `order` option to the finder. See https://github.com/typeorm/typeorm/blob/master/docs/find-options.md

src/organisations/helpers/organisations-filter.helper.ts
```diff
@@ -0,0 +1,43 @@
+import { BaseFilterHelper } from '../../helpers/base-filter.helper';
+import { Industry } from '../../industries/industry.entity';
+import { Organisation } from '../organisation.entity';
+
+export class OrganisationsFilterHelper extends BaseFilterHelper<Organisation> {
```
pezholio, line 5: I've completely missed that we're doing this at the code level, which may not be the best way of approaching this. In the spirit of getting this out, I think it's fine to keep this as is now, with a view to taking this down to the database layer later.

pezholio (commented):
Looks good - I am minded to at least shift the sorting to the database level. I think we should be doing the searching in the database too ideally, but I think it's fine to keep this for now in the spirit of rapid iteration 👍

Gweaton (approved):
Just a couple of tiny things in addition to Stu's points.

src/organisations/organisations.service.spec.ts
```diff
@@ -36,22 +36,46 @@ describe('OrganisationsService', () => {
...
+    it('returns all Organisations, populated with Professions', async () => {
       expect(organisations).toEqual([organisation]);
-      expect(repoSpy).toHaveBeenCalledWith({ relations: ['professions'] });
+      expect(repoSpy.mock.calls[0][0]).toMatchObject({
+        relations: ['professions'],
+      });
+    });
+
+    it('Organisations should be sorted', () => {
+      expect(repoSpy.mock.calls[0][0]).toMatchObject({
```
Gweaton, line 76: Rather than `calls[0][0]` we can put this, which is a bit more readable:
```
      expect(repoSpy).toHaveBeenCalledWith(
        expect.objectContaining({
          order: { name: 'ASC' },
        }),
```

src/professions/professions.service.spec.ts
```diff
@@ -58,22 +58,46 @@ describe('Profession', () => {
...
 
+    beforeEach(async () => {
+      repoSpy = jest.spyOn(repo, 'find');
+      professions = await service.allConfirmed();
+    });
+
+    it('should return all confirmed Professions', async () => {
       expect(professions).toEqual(professionArray);
-      expect(repoSpy).toHaveBeenCalledWith({ where: { confirmed: true } });
+      expect(repoSpy.mock.calls[0][0]).toMatchObject({
```
Gweaton, line 92: see comment on not using `calls` above.

Gweaton (approved):
This looks good, have just suggested a way of making tests a bit more readable.

PR #114, opened 2022-01-13
1942 Show Regulatory Authorities to an admin

pezholio:
This builds on #111 to add a view of a regulatory authority on the admin side. The buttons to edit and delete are currently not functional, but will be sorted in th next bit of work. 

# Screenshots 

image

### DIFF
--- a/assets/stylesheets/application.scss
+++ b/assets/stylesheets/application.scss
@@ -3,95 +3,92 @@ $govuk-assets-path: '~govuk-frontend/govuk/assets/';
 @import 'govuk/all';
 @import './admin.scss';
 
-.rpr-details__section-title {
-  text-transform: uppercase;
-}
-
-.rpr-listing__filters-container {
-  background-color: govuk-colour('light-grey');
-  padding: 1rem;
-}
-
-.rpr-listing__filter {
-  background-color: govuk-colour('white');
-  height: 180px;
-  overflow-y: auto;
-}
-
-.rpr-listing__search-criteria-container {
-  margin-bottom: 5px;
-}
-
-.rpr-listing__profession-title:link,
-.rpr-listing__ra-title:link {
-  text-decoration: none;
-}
-
-.rpr-listing__profession-title:hover,
-.rpr-listing__ra-title:hover {
-  text-decoration: underline;
-}
-
-.rpr-listing__profession-summary-list dt,
-.rpr-listing__profession-summary-list dd,
-.rpr-listing__ra-summary-list dt,
-.rpr-listing__ra-summary-list dd {
-  padding: 2px 0;
-}
-
-.rpr-internal__page {
-  .govuk-header__container {
-    border-bottom: 10px solid govuk-colour('yellow');
-  }
-}
-
-.rpr-listing__filters-container {
-  background-color: govuk-colour('light-grey');
-  padding: 1rem;
-}
-
-.rpr-listing__filter {
-  background-color: govuk-colour('white');
-  height: 180px;
-  overflow-y: auto;
-}
-
-.rpr-internal__listing {
-  .govuk-width-container {
-    max-width: 1200px;
-  }
-}
-
-.rpr-internal-listing__filters-container {
-  background-color: govuk-colour('light-grey');
-  padding: 1rem;
-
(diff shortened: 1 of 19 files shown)


assets/stylesheets/application.scss
```diff
@@ -3,95 +3,103 @@ $govuk-assets-path: '~govuk-frontend/govuk/assets/';
...
+.rpr-details {
+  &__section-title {
+    text-transform: uppercase;
+  }
 
-.rpr-listing__filter {
-  background-color: govuk-colour('white');
-  height: 180px;
-  overflow-y: auto;
+  &__section-title {
```
Gweaton, line 16: Is this a duplicate of the one above, or am I missing something?

assets/stylesheets/application.scss
```diff
@@ -3,95 +3,103 @@ $govuk-assets-path: '~govuk-frontend/govuk/assets/';
...
+  &__ra-summary-list dd {
+    padding: 2px 0;
+  }
 
-.rpr-listing__filter {
-  background-color: govuk-colour('white');
-  height: 180px;
-  overflow-y: auto;
-}
+  &__filters-container {
```
Gweaton, line 54: again, is this a duplicate of line 23 here?

assets/stylesheets/application.scss
```diff
@@ -3,95 +3,103 @@ $govuk-assets-path: '~govuk-frontend/govuk/assets/';
...
+    background-color: govuk-colour('light-grey');
+    padding: 1rem;
+  }
 
-.rpr-internal__listing {
-  .govuk-width-container {
-    max-width: 1200px;
+  &__filter {
```
Gweaton, line 59: Duplicate of line 27?

src/organisations/admin/organisations.controller.spec.ts
```diff
@@ -31,21 +42,55 @@ jest.mock('../presenters/organisations.presenter', () => {
   };
 });
 
+jest.mock('../presenters/organisation.presenter', () => {
```
Gweaton, line 45: Interesting, I quite like this approach of mocking out external classes - do you think this gives us enough integration coverage? I suppose this is probably covered in the e2e tests, right?

src/organisations/presenters/organisation.presenter.ts
```diff
@@ -20,7 +20,12 @@ export class OrganisationPresenter {
         html: await this.industries(),
       },
       {
-        text: '',
+        html: `<a class="govuk-link" href="/admin/organisations/${this.organisation.slug}">
+          View details
+          <span class="govuk-visually-hidden">
+            about ${this.organisation.name}
```
Gweaton, line 26: Something I've been thinking about here is, we should probably be escaping all these user-entered values when we render HTML like this. We do it in a lot of places, but I think it'd be good to have a piece of work to tidy it up where relevant.

Gweaton (approved):
Looks good, a few minor things.

assets/stylesheets/application.scss
pezholio: Good catch 👍

src/organisations/admin/organisations.controller.spec.ts
pezholio: Yeah, it's also coverted by the presenter tests. Trying to be as solitary as possible here 👍

src/organisations/presenters/organisation.presenter.ts
pezholio: That's a good shout. I don't think we get anything for free in the JS standardlib, so I'll add a card to tidy this up later 👍

Gweaton: Do you have access to view [this PR](https://github.com/ministryofjustice/hmpps-interventions-ui/commit/8995de8b62e3ced7b34fede10833df35df6a84ab)? Where we're interpolating stuff in templates, we can use the `escape` param, but if we're constructing HTML in a presenter, we could use this. I suppose that gives us the benefit of escaping consistently throughout the service.

pezholio: Ah, nice 👍

PR #116, opened 2022-01-14
(1985) Add page for setting Legislation information on a Profession

Gweaton:
# Changes in this PR

Adds final page to "Add a profession" journey for adding Legislation information. This creates a new Legislation on the Profession.

We initially planned to store multiple Legislations on the Profession, but it looks like this was in error, so we've added a single "one-to-one" field here. We'll remove the additional field once this is merged in.

### Legislation page

image

#### Legislation page with errors

image

### Check your answers

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - List Regulatory Authorities
 - Add page for adding a new qualification to a profession
 - Show Regulatory Authorities to an admin
+- Add page for setting legislation information on a profession
 
 ### Changed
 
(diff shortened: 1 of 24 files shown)


bryangaledxw (approved):
Looks good. The thing I want to understand though is where do we use the generic `backLink` in `common/utils.ts`, and where we need to specify that more manually?

Gweaton:
> Looks good. The thing I want to understand though is where do we use the generic `backLink` in `common/utils.ts`, and where we need to specify that more manually?

@bryangaledxw It's a good question. I initially tried to use the generic backlink, but on pages where we have a POST request that posts to a different route than the render route e.g. GET `legislation/edit` renders the page but POST `legislation` creates the legislation, if there's a validation error then the user resubmits the form, the generic backlink render that POST route (as that's the referrer).

This isn't ideal, as most of the pages we have backlinks on submit some kind of data, but I think it's best to try and conditionally hardcode the backlink like I have throughout this journey where there are POST requests that POST to a different route than the the one being rendered, otherwise it's fine to use the generic back link.

PR #118, opened 2022-01-17
Redirect user to `/admin` after login

pezholio:
Before, the user was getting redirected to the root url, which was annoying / unexpected. This redirects to the `/admin` route, which is more expected.


bryangaledxw (approved):
🥳

Super! This was annoying me!

Gweaton (approved):

pezholio:
@bryangaledxw Yeah, I thought we should sort it before we start user testing!

PR #127, opened 2022-01-19
Tidy up add profession views

Gweaton:
# Changes in this PR

Adds yello banner to add / edit professions to make it clearer which version of the site (public / internal) users are viewing.

### Before

image

### After

image


pezholio (approved):
LGTM - pending a rebase

PR #128, opened 2022-01-19
(2023) Remove placeholder text on Professions

Gweaton:
# Changes in this PR

- Displays real Organisation on a Profession page, rather than placeholder text
- Fixes typo
- Removes duplicate "mandatory professional experience" answer
- Displays human-readable qualification answers, rather than ugly raw data values

### Before

image

### After

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Update start page
 - Add public-facing Regulatory Authority pages
 - Allow Professions to be edited
+- Display real data on "view profession" pages
 
 ### Changed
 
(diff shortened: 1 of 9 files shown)


bryangaledxw (approved):
Looks good. Might be more than you want to do in this PR, but feels like `professions.form.radioButtons.methodsToObtainQualification` is the wrong namespace now for the qualification translations, given `QualificationPresenter` is used outside of forms. Could be just `professions.methodsToObtainQualification`?

Gweaton:
> Looks good. Might be more than you want to do in this PR, but feels like `professions.form.radioButtons.methodsToObtainQualification` is the wrong namespace now for the qualification translations, given `QualificationPresenter` is used outside of forms. Could be just `professions.methodsToObtainQualification`?

@bryangaledxw good point. I'm going to merge this now so it's there for user testing in the morning, but will update this in its own PR.

PR #129, opened 2022-01-19
Filter unconfirmed professions in `OrganisationService`

bryangaledxw:
# Changes in this PR

When requesting Organisations with the professions relation from `OrganisationService`, remove unconfirmed Professions. This fixes the issue currently visible on staging, where trying to view a profession with an unconfirmed Profession gives an error

### DIFF
--- a/src/organisations/organisations.service.spec.ts
+++ b/src/organisations/organisations.service.spec.ts
@@ -91,9 +91,9 @@ describe('OrganisationsService', () => {
   describe('find', () => {
     it('returns an Organisation', async () => {
       const repoSpy = jest.spyOn(repo, 'findOne');
-      const organisations = await service.find('some-uuid');
+      const foundOrganisation = await service.find('some-uuid');
 
-      expect(organisations).toEqual(organisation);
+      expect(foundOrganisation).toEqual(organisation);
       expect(repoSpy).toHaveBeenCalled();
     });
   });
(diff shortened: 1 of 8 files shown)


src/organisations/organisations.service.spec.ts
```diff
@@ -2,30 +2,38 @@ import { Test, TestingModule } from '@nestjs/testing';
...
+  const unconfirmedProfession = professionFactory.build({ confirmed: false });
+
+  const organisation = organisationFactory.build({
+    professions: [confirmedProfession, unconfirmedProfession],
+  });
 
   beforeEach(async () => {
     const module: TestingModule = await Test.createTestingModule({
       providers: [
+        // Rebuild organisations on each call to our mock repository, as they
```
Gweaton, line 23: Internally at dxw we've recently (in the last week) been talking about how we can standardise the way we write our tests. One idea we've decided on is to handle as much of the test setup as possible in each individual `it` block, rather than in a `beforeEach`. This means context is contained to an indivual test and means readers don't have to scroll up and down the file to see expectations e.g. objects set up at the top of the file.

This is obviously not the way we originally wrote our tests, so there's some refactoring we'll probably want to do gradually, but I think it'd be good to start moving in that direction for all new code, or when adding new tests.

So in this suite we'd stub a resolved value for the repository `find` function in each test.

Does that seem sensible to you?

We can do this in a separate PR if you'd rather, so we get the fix out sooner.

Gweaton (approved):
Looks good. Just one comment that we can revisit once this is merged if you agree.

PR #130, opened 2022-01-19
Remove ID column from admin professions display

bryangaledxw:
# Changes in this PR

Remove ID column from admin professions display

### Before

localhost_3000_admin_professions

### After

localhost_3000_admin_professions

### DIFF
--- a/src/professions/admin/list-entry.presenter.spec.ts
+++ b/src/professions/admin/list-entry.presenter.spec.ts
@@ -30,7 +30,6 @@ describe('ListEntryPresenter', () => {
 
       const expected: TableRow = [
         { text: 'Example Profession' },
-        { text: 'PLCH0LD3R' },
         {
           text: `${translationOf('nations.scotland')}, ${translationOf(
             'nations.northernIreland',
@@ -76,7 +75,6 @@ describe('ListEntryPresenter', () => {
 
       const expected: TableRow = [
         { text: 'Example Profession' },
-        { text: 'PLCH0LD3R' },
         {
           text: `${translationOf('nations.scotland')}, ${translationOf(
             'nations.northernIreland',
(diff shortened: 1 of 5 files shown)


src/professions/admin/list-entry.presenter.spec.ts
```diff
@@ -113,8 +108,37 @@ describe('ListEntryPresenter', () => {
...
+
+function createPresenter(profession: Profession): ListEntryPresenter {
+  const i18nService = createPopulatedI18nService();
+
+  return new ListEntryPresenter(profession, i18nService);
+}
+
+function createPopulatedI18nService(): DeepMocked<I18nService> {
+  return createMockI18nService({
+    'nations.scotland': 'Scotland',
```
Gweaton, line 124: @pezholio made a [fair point on a recent PR](https://github.com/UKGovernmentBEIS/regulated-professions-register/pull/115#discussion_r784662301) of mine that we don't necessarily need to worry about the translated value here, and that this setup I followed his example of asserting that the text value is "Translation of `industries.law`" etc.

I suppose if we're thinking of this being a "unit" test, we don't really need to test what the I8nService is doing, just that it's returning a value that we expect.

Obviously asserting on "translation of x" every time isn't ideal either, so would be keen to hear your thoughts on this too.

I suppose if we were to remove this setup, we probably wouldn't need this helper, but if we want to keep it, I think the structure of these reusable functions is sensible and readable.

src/professions/admin/list-entry.presenter.spec.ts
```diff
@@ -113,8 +108,37 @@ describe('ListEntryPresenter', () => {
...
+        ListEntryPresenter.headings(
+          createPopulatedI18nService(),
+          'single-organisation',
+        ),
       ).resolves.toEqual(expected);
     });
   });
 });
+
+function createPresenter(profession: Profession): ListEntryPresenter {
```
Gweaton, line 116: We could possibly just create the `new ListEntryPresenter(profession, createPopulatedI18nService()` in the test body itself, but I also think it's quite nice to keep it self-contained here, particularly if it were to get more complex in future.

Gweaton (approved):
I think this seems sensible, have just added a couple of comments for potential discussion.

Gweaton:
[Here's an article](https://thoughtbot.com/blog/lets-not) I read a few years ago about structuring tests - it references Ruby and RSpec but the principles are the same.

PR #131, opened 2022-01-19
(2021) Make the details pages print-friendly

mingoland:
# Changes in this PR
Adding CSS so that:
- phase banner
- back button
- sidebar buttons
on profession's details page and regulator's details page (both public facing and internal) will be invisible when users print these pages.

### Before
Phase banner and back link in the top area:
Screenshot 2022-01-19 at 18 02 24
Buttons at the sidebar:
Screenshot 2022-01-19 at 18 02 40

### After
Phase banner and back link removed:
Screenshot 2022-01-19 at 18 03 04
Buttons at the sidebar removed:
Screenshot 2022-01-19 at 18 03 49


pezholio (approved):
LGTM 👍

PR #134, opened 2022-01-20
(2035) Link to internal professions page from internal organisations page

Gweaton:
# Changes in this PR

- Make Profession link from `admin/organisation/:slug` take users to `admin/profession/:slug` rather than public-facing profession page.
- Add back link to internal `admin/profession/:slug` page to take users to index page


pezholio (approved):
LGTM 👍

PR #135, opened 2022-01-20
(1977) Move back button to under phase banner

pezholio:
This moves the back button to under the phase banner to avoid the awkward spacing issues we have had and also DRY up the templates. I've also added a `BackLink` decorator, which sets the `backLink` without it having to be present in templates. 

### Before

image

### After

image

### DIFF
--- a/src/common/back-link.interceptor.spec.ts
+++ b/src/common/back-link.interceptor.spec.ts
@@ -1,0 +1,88 @@
+import { ExecutionContext, CallHandler } from '@nestjs/common';
+import { BackLinkInterceptor } from './back-link.interceptor';
+import { createMock, DeepMocked } from '@golevelup/ts-jest';
+import { of } from 'rxjs';
+
+describe('BackLinkInterceptor', () => {
+  let context: DeepMocked<ExecutionContext>;
+  let next: DeepMocked<CallHandler>;
+
+  beforeEach(() => {
+    data = { foo: 'bar' };
+    context = createMock<ExecutionContext>();
+    next = createMock<CallHandler>({
+      handle: jest.fn(() => of(data)),
+    });
+  });
+
+  it('should add the backLink to the call handler', () => {
+    const backLink = '/foo/bar';
+    const interceptor = new BackLinkInterceptor(backLink);
+
+    const response = interceptor.intercept(context, next);
+
+    response.subscribe({
+      next: (value) => {
+        expect(value).toEqual({ ...data, backLink: backLink });
+      },
+    });
+  });
+
+  describe('when there are params specified', () => {
+    beforeEach(() => {
+      context = createMock<ExecutionContext>({
+        switchToHttp: () => ({
+          getRequest: () => {
+            return {
+              params: {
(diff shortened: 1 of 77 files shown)


bryangaledxw:
Very happy about the back link decorator!

src/professions/admin/professions.presenter.ts
```diff
@@ -77,7 +76,6 @@ export class ProfessionsPresenter {
         ),
         changedBy: [],
       },
-      backLink: backLink(this.request),
```
bryangaledxw, line 80: `request` is no longer needed in the constructor

bryangaledxw (approved):
Looks good!

src/common/back-link.interceptor.ts
```diff
@@ -0,0 +1,56 @@
...
+import { Request } from 'express';
+
+export interface Response<T> {
+  data: T;
+}
+
+export type Generator = (request: Request) => string;
+@Injectable()
+export class BackLinkInterceptor<T> implements NestInterceptor<T, Response<T>> {
+  backLink: string;
```
bryangaledxw, line 17: These can be `private readonly`

pezholio:
Ta for the review! 👍

src/common/back-link.interceptor.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+  let data: any;
+  let context: DeepMocked<ExecutionContext>;
+  let next: DeepMocked<CallHandler>;
+  let request: DeepMocked<Request>;
+  let response: DeepMocked<Response>;
+
+  beforeEach(() => {
+    data = { foo: 'bar' };
+    next = createMock<CallHandler>({
+      handle: jest.fn(() => of(data)),
```
bryangaledxw, line 17: What does `of` get us here? I'm looking at the docs for it and I feel like I'm not getting it!

src/common/back-link.interceptor.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+
+describe('BackLinkInterceptor', () => {
+  let data: any;
+  let context: DeepMocked<ExecutionContext>;
+  let next: DeepMocked<CallHandler>;
+  let request: DeepMocked<Request>;
+  let response: DeepMocked<Response>;
+
+  beforeEach(() => {
+    data = { foo: 'bar' };
```
bryangaledxw, line 15: Can we do `const data` here and remove the above declaration? Or just move it inline?

bryangaledxw (approved):

src/common/back-link.interceptor.spec.ts
pezholio: We don't even need it now, was an artifact of something else I was trying

src/common/back-link.interceptor.spec.ts
pezholio: That was basically cargo-culted from somewhere - I can't remember what I was trying to do! It's not necessary now anyway, so I've removed it 👍

PR #136, opened 2022-01-21
Add Dependabot

pezholio:
I thought this was automatic, but apparently not!


Gweaton (approved):
Yeah, I just updated govuk frontend in mine so was a bit confused that we hadn't been prompted!

PR #137, opened 2022-01-21
(2036) Add an OrganisationsVersions table

pezholio:
This adds an `OrganisationVersions` entity and service, which is the first part of the puzzle for versioning and making draft versions of our Organisations.

### DIFF
--- a/src/organisations/organisation-version.entity.ts
+++ b/src/organisations/organisation-version.entity.ts
@@ -1,0 +1,56 @@
+import { Organisation } from './organisation.entity';
+import { User } from '../users/user.entity';
+import {
+  Entity,
+  PrimaryGeneratedColumn,
+  Column,
+  CreateDateColumn,
+  UpdateDateColumn,
+  Index,
+  OneToOne,
+  JoinColumn,
+  ManyToOne,
+} from 'typeorm';
+
+export enum OrganisationVersionStatus {
+  Live = 'live',
+  Draft = 'draft',
+  Archived = 'archived',
+}
+
+@Entity({ name: 'organisationVersions' })
+export class OrganisationVersion {
+  @PrimaryGeneratedColumn('uuid')
+  id: string;
+
+  @Index({ unique: true, where: '"slug" IS NOT NULL' })
+  @Column({ nullable: true })
+  slug: string;
+
+  @OneToOne(() => Organisation, (organisation) => organisation.version)
+  @JoinColumn()
+  organisation: Organisation;
+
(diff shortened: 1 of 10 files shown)


src/organisations/organisation-version.entity.ts
```diff
@@ -0,0 +1,56 @@
...
+@Entity({ name: 'organisationVersions' })
+export class OrganisationVersion {
+  @PrimaryGeneratedColumn('uuid')
+  id: string;
+
+  @Index({ unique: true, where: '"slug" IS NOT NULL' })
+  @Column({ nullable: true })
+  slug: string;
+
+  @OneToOne(() => Organisation, (organisation) => organisation.version)
```
Gweaton, line 30: I may need to re-read the ADR for this, but do we ever think we might need to keep track of multiple versions of an organisation e.g. for some kind of audit trail?

src/organisations/organisation-versions.service.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+  let repo: Repository<OrganisationVersion>;
+
+  beforeEach(async () => {
+    const module: TestingModule = await Test.createTestingModule({
+      providers: [
+        OrganisationVersionsService,
+        {
+          provide: getRepositoryToken(OrganisationVersion),
+          useValue: {
+            save: (organisationVersion: OrganisationVersion) => {
```
Gweaton, line 19: I know we're following a pattern from previous services here, but after the testing standards discussion, I think we want to move test setup into individual `it` blocks as much as possible, rather than the `beforeEach`, to keep test context contained to the immediate test so readers don't have to scroll up and down to figure out what `organisationVersion` is?

So here we'd move this `save` stubbing here into the `it` block? So in the `beforeEach` we'd just set up a blank, mock `OrganisationVersion` each time, without stubbing any return values.

Gweaton (approved):
Thanks for sorting this, have a couple of comments/questions but nothing major

src/organisations/organisation-version.entity.ts
pezholio: I've been thinking about this over the weekend (sad I know), but I'm beginning to think we might want to add a field to our organisations / professions to add a parent ID field, so we can do just that

src/organisations/organisation-versions.service.spec.ts
pezholio: I mean, that said, we don't even need to set up a value here, because we're stubbing the whole method

PR #143, opened 2022-01-24
Change our exception filter to report to Rollbar

pezholio:
Our HttpExceptionFilter was only catching HTTP exceptions (as the name, I guess suggests!), so any errors that weren't standard HTTP errors were not getting caught by this filter.

I've tweaked it to now accept all kind of errors, catch the standard HTTP ones, and show a friendly error page and log the error. If we're in production, we also report the error to Rollbar, as the standard Express Rollbar middleware wasn't working as expected


src/common/global-exception.filter.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+        expect(reponse.render).toHaveBeenCalledWith('errors/generic-error');
+      });
+
+      describe('when NODE_ENV is not set to production', () => {
+        beforeEach(() => {
+          process.env['NODE_ENV'] = 'test';
+        });
+
+        it('should report the error to rollbar', () => {
+          service.catch(exception, host);
```
Gweaton, line 175: Again, do we want to set a new `exception` object in each test here, or create it in a `beforeEach`? Thinking about our testing standards, we centered on creating the test object in the `it`, so context was contained in the text.

This is a simple test though, and it's nicely grouped into `describe`s - I just wonder if we need to settle on a consistent way of doing things here. For me, `service.catch(exception, host)` feels a bit opaque here, as I have to scroll up to figure out if `exception` or `host` is what's causing the error, but it's actually `NODE_ENV`. Does that make any sense?

src/common/global-exception.filter.ts
```diff
@@ -0,0 +1,74 @@
...
+      case HttpStatus.FORBIDDEN: {
+        response.render('errors/forbidden');
+        break;
+      }
+      case HttpStatus.NOT_FOUND: {
+        response.render('errors/not-found');
+        break;
+      }
+      default: {
+        if (process.env['NODE_ENV'] === 'production') {
```
Gweaton, line 48: Can I just double check for my own understanding here - does `production` here mean Production and Staging? If it does, I wonder if we should wrap this in a clearly-named method to make that obvious.

Gweaton (approved):
A couple of small things

PR #145, opened 2022-01-24
(1986) Add Cypress-Axe to e2e tests and fix accessibility issues

Gweaton:
# Changes in this PR

- Uses Cypress-Axe as much as possible in our e2e tests to check for obvious accessibility issues
- Fixes these issues where possible, or ignores them where the govuk components don't provide the level of customisation required to fix them (where not a critical issue).


bryangaledxw (approved):
Looks great!

For the two places where we have

```cy.injectAxe();
      cy.checkA11y(
        {
          exclude: [['#phase-banner-container']],
        },
        {
          // Conditional radio buttons add an additional `aria-expanded` field,
          // but unfortunately we can't disable the rule for a specific element,
          // so we're not using the usual helper here and disabling that rule
          rules: { 'aria-allowed-attr': { enabled: false } },
        },
      );
```
is it worth letting `cy.checkAccessibility()` take an optional param so we could do something like:

`cy.checkAccessibility({ 'aria-allowed-attr': { enabled: false } })`

PR #149, opened 2022-01-24
(2018) Internal RA listing may be filtered

bryangaledxw:
# Changes in this PR

Adds filters to the internal RA listing

### Before
localhost_3000_admin_organisations (1)

### After
localhost_3000_admin_organisations


pezholio (approved):
LGTM - thanks for this 👍  I am still a bit worried about us doing this filtering with code, but I think this is fine for now

PR #150, opened 2022-01-24
Update internal `OrganisationsController` tests

bryangaledxw:
This reduces the reliance on `beforeEach`, and makes more explicit what test depend on which `OrganisationsService` methods.

I've kept the creation of the controller itself inside the `beforeEach`, which seems reasonable for controller tests, and I think it makes sense to keep `i18nService` and `organisationsService` as variables in the top level `describe`, rather than each `it` to avoid a lot of repetition.

I'd maybe want to move the `jest.mock()` module mocking calls into the `it` blocks, but this doesn't seem possible without everything breaking, and it makes some sense to have these just after the imports


pezholio (approved):
This is so much better - thank you! 🙌

PR #151, opened 2022-01-25
Iterate our approach to versioning

pezholio:
Sincestarting to implement versioing, we've encountered a problem where copying entities will mean we will no longer be able to index slugs uniquely, causing us problems when we try and create new entities and fetch entities by slug.

I've added a new ADR to capture an iterated way of handling versioning, which hopefully solves this issue and also makes some other stuff a bit tidier too.


Floppy (approved):
Yep, seems reasonable, the slug should remain constant and doesn't need versioning.

doc/architecture/decisions/0015-use-versions-to-store-versioned-entity-data.md
```diff
@@ -0,0 +1,43 @@
...
+Because Organisations are created on one page, we can create the Organisation table and
+the Version at the same time
+
+## Consequences
+
+This means we will be able to uniquely index on slugs without having to remove the index
+so we can copy entities. It will also make it much easier to get the latest version for an
+entity (for example, they can have a `#latest_version` and `#latest_draft` method).
+
+This does however mean that we will have to make sure that names cannot be edited once they
```
Gweaton, line 38: I think this makes sense, if slugs need to be immutable - we'll want to make sure we have wording/instructions in the service that admins/BEIS will need to create a new organisation / profession instead of renaming it.

doc/architecture/diagrams/versioning/flow-2_0.wsd
```diff
@@ -0,0 +1,33 @@
+@startuml flow
+(*) --> "Create new Profession"
+--> "Add basic properties"
```
Gweaton, line 3: So are these basic properties really just the name at this stage? Just wondering what happens at this stage and what happens at the "Form is completed" stage. We might want to rejig the struture of the profession journey e.g. the Top Level Details page to just have a simple "name entry" page. 

We know this journey will probably need tidying up anyway after user research and with different users from different organisations entering different information, so I don't think it's a big deal.

doc/architecture/diagrams/versioning/flow-2_0.wsd
```diff
@@ -0,0 +1,33 @@
...
+--> "Version is created"
+--> "Organisation is confirmed"
+--> "Organisation is created with a unique slug"
+--> "Version is marked as a draft"
+
+(*) --> "Entity is edited"
+--> if "A draft exists" then
+  -->[Yes] "Edit existing draft"
+else
+  -->[No] "Copy existing Entity"
```
Gweaton, line 21: Is this actually a step, or does Entity here refer to the `Version`? Could this step and the next be combined into one? When first reading this, I thought it meant we'd be copying the existing `Profession`/`Organisation`, but we'll be using the same top-level entity, right?

Gweaton (approved):
I think this is solid - I like that we'll have consistent slugs, and having versions store most of the data means we'll be able to compare versions more easily against oneanother e.g. if there's a future requirement to see who updated which field(s).

Just have a couple of clarifications in comments.

doc/architecture/diagrams/versioning/versions-2_0.wsd
```diff
@@ -0,0 +1,49 @@
+@startuml versions
+
+entity OrganisationVersion {
+  Represents a version of an Organisation
+  --
+  * id: uuid
+  * status: enum (live/draft/archived)
```
bryangaledxw, line 7: Again thinking of what happens on the final step of the draft editing flow, I wonder if this should be "live/draft/archived/unconfirmed". So until the user hits the final confirm button, they're editing an unconfirmed version, and then when they hit that button it gets promoted to draft and the old draft gets deleted/archived.

Then we also have an easy rule with any editing that we check the ID in the url is of an unconfirmed entity

bryangaledxw (approved):

doc/architecture/diagrams/versioning/flow-2_0.wsd
bryangaledxw: Yep, seconding that I'm unsure what entity means here, when I think this step wants to be unambiguous which specific table we're copying a row of

doc/architecture/diagrams/versioning/flow-2_0.wsd
pezholio: Yeah, I think we'll maybe have the organisation on the first page too I reckon

doc/architecture/diagrams/versioning/flow-2_0.wsd
pezholio: Yeah, sorry, that was an artefact of the older version. I've tweaked this now, so it's clearer

pezholio:
Thanks for the comments everyone. I've made a few tweaks and set this to automerge 👍

PR #152, opened 2022-01-25
Allow organisations to be created

pezholio:
This allows organisations to be created. The journey isn't 100% complete (for example, we don't create a slug), but there didn't seem much point in adding that given that we're going to make some quite significant changes to the model soon (given the changes in #151).

There are some copy changes that need making (for example, it still says "amending" rather than "creating" on the forms), but this can be tightened up once we start with the versioning.

image

image

image

image

### DIFF
--- a/src/organisations/admin/organisations.controller.ts
+++ b/src/organisations/admin/organisations.controller.ts
@@ -4,13 +4,14 @@ import {
   Get,
   Render,
   Param,
-  Post,
   Body,
   UseFilters,
   Put,
   Query,
+  Res,
 } from '@nestjs/common';
 import { I18nService } from 'nestjs-i18n';
+import { Response } from 'express';
 
 import { AuthenticationGuard } from '../../common/authentication.guard';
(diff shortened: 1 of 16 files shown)


src/organisations/admin/organisations.controller.ts
```diff
@@ -62,24 +62,24 @@ export class OrganisationsController {
     return organisationSummaryPresenter.present();
   }
 
-  @Get('/:slug/edit')
+  @Get('/:id/edit')
```
Gweaton, line 84: For the profession journey, we used the `slug` for edit, but then switch to the `id` after that - should we update that to use the `id` there instead too?

Gweaton (approved):

src/organisations/admin/organisations.controller.ts
pezholio: Yeah, I think so. This is all stuff we can tighten up as we go though

PR #156, opened 2022-01-25
(2057) Fix admin view where no qualifications are found for a profession

Gweaton:
# Changes in this PR

- Ensures we don't create a `QualificationPresenter` with `null` when the Profession has no Qualification. This was causing the 'view details' link to break on staging.
- Tidies up tests a bit


pezholio (approved):
LGTM 👍

PR #159, opened 2022-01-25
(2007) HTML escaping

bryangaledxw:
# Changes in this PR

Fix instances where HTML entered by the user can be presented back unescaped

### Before
localhost_3000_professions_registered-trademark-attorney

### After
localhost_3000_professions_registered-trademark-attorney (1)

### DIFF
--- a/src/organisations/presenters/organisation.presenter.spec.ts
+++ b/src/organisations/presenters/organisation.presenter.spec.ts
@@ -10,6 +10,10 @@ import { createMockI18nService } from '../../testutils/create-mock-i18n-service'
 import organisationFactory from '../../testutils/factories/organisation';
 import professionFactory from '../../testutils/factories/profession';
 import industryFactory from '../../testutils/factories/industry';
+import { escape } from '../../helpers/escape.helper';
+import { escapeOf } from '../../testutils/escape-of';
+
+jest.mock('../../helpers/escape.helper');
 
 describe('OrganisationPresenter', () => {
   let organisation: Organisation;
@@ -34,6 +38,8 @@ describe('OrganisationPresenter', () => {
         });
 
         it('returns the table row data', async () => {
+          (escape as jest.Mock).mockImplementation(escapeOf);
+
           const presenter = new OrganisationPresenter(
             organisation,
             i18nService,
@@ -51,8 +57,12 @@ describe('OrganisationPresenter', () => {
             ),
           });
           expect(tableRow[3]).toEqual({
-            html: expect.stringContaining(`about ${organisation.name}`),
+            html: expect.stringContaining(
+              `about ${escapeOf(organisation.name)}`,
+            ),
           });
+
+          expect(escape).toBeCalledWith(organisation.name);
         });
       });
 
@@ -77,6 +87,8 @@ describe('OrganisationPresenter', () => {
         });
 
         it('returns the table row data', async () => {
+          (escape as jest.Mock).mockImplementation(escapeOf);
+
           const presenter = new OrganisationPresenter(
             organisation,
             i18nService,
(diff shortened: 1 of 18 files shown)


src/users/user.presenter.spec.ts
```diff
@@ -53,46 +44,58 @@ describe('UserPresenter', () => {
...
+    name: 'name',
+    email: 'email@example.com',
+    externalIdentifier: '212121',
+    permissions: permissions,
+  });
+
+  return user;
+}
+
+function createPresenter(user: User): UserPresenter {
```
Gweaton, line 117: I'd argue that this probably doesn't need to be wrapped in a helper function for two reasons:
- As this is the `UserPresenter` test, I find a bit more clear to see the `new ObjectBeingTested()` as close as possible to the top of the `it` block as that's the object we care about. Wrapping it in a helper suggests there could be some more complex setup hidden away that I have to scroll down for.
- It's simple setup, and wrapping it in a helper doesn't save us a lot of lines of code.

For all other collaborators, or for more complex setup it makes sense to encapsulate them in reusable helpers, but I think this is simple enough not to need a helper here.

views/admin/professions/check-your-answers.njk
```diff
@@ -58,7 +58,7 @@
                   text: ("professions.form.label.topLevelInformation.nations" | t)
                 },
                 value: {
-                  html: selectedNationsSummaryListHtml | safe
```
Gweaton, line 61: I think I foolishly assumed `safe` was the same as `escape` 😬. Well done for sorting this.

src/users/user.presenter.spec.ts
```diff
@@ -40,6 +40,24 @@ describe('UserPresenter', () => {
     `.replace(/(\n)/gm, '');
       expect(presenter.showLink().replace(/(\n)/gm, '')).toEqual(expected);
     });
+
+    it("should escape the user's name", () => {
```
Gweaton, line 44: Because we have this behaviour tested in the `escape` specs, I wonder if we can just assert that `escape` was called here when we call `showLink`?

src/organisations/presenters/organisation.presenter.spec.ts
```diff
@@ -294,6 +294,19 @@ describe('OrganisationPresenter', () => {
...
+
+    it('escapes the address', () => {
+      organisation = organisationFactory.build({
+        address:
+          '<b>Bold Street</b>, London, E14 5AB<script src="bad.js"></script>',
+      });
+
+      const presenter = new OrganisationPresenter(organisation, i18nService);
+
+      expect(presenter.address()).toEqual(
```
Gweaton, line 306: Again, do you think it makes sense to check `escape` was called here? If you make `escape` a mock function and assert that it was called, I think that would make test setup a bit easier here.

Gweaton (approved):
Thanks for sorting this. The implementation looks good and it's good to have a pattern to follow here.

I just had a couple of comments about testing, really - I think for me the big one is "what do we test in a unit test?" Do we want to test the output of other classes, or do we trust that our tests for the other class are reliable and just make sure we call that method?

src/users/user.presenter.spec.ts
bryangaledxw: Yep, makes sense!

src/users/user.presenter.spec.ts
bryangaledxw: Yep, makes sense. Will do something similar to `translationOf`

src/organisations/presenters/organisation.presenter.spec.ts
Gweaton: Nice, looks good, thanks for sorting.

PR #160, opened 2022-01-26
(2036) Allow versioning of Organisations

pezholio:
This moves all the data fields that were already present in an Organisation to an OrganisationVersion. When an Organisation is first created, a version is created alongside that organisation. Once the organisation is confirmed, the version is set to be
a draft.

Apologies for the enormous PR - I was hoping to do it in a piecemeal fashion, but altering the model meant a whole bunch of other things broke! I've skipped a few e2e tests for now - I'll put these in in the next PR once this has been done!

### DIFF
--- a/cypress/integration/admin/organisations/edit.spec.ts
+++ b/cypress/integration/admin/organisations/edit.spec.ts
@@ -1,4 +1,4 @@
-describe('Editing organisations', () => {
+describe.skip('Editing organisations', () => {
   context('When I am logged in as admin', () => {
     beforeEach(() => {
       cy.loginAuth0('admin');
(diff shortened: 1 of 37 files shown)


Gweaton (commented):
There's a lot going on here, but I think the approach is sound, and a good pattern to follow. I can't see anything that worries me, so think this is good to go.

I'm hesitant to merge this as-is though, as it breaks a lot of functionality on staging. I wonder if it's worth opening all the PRs in sequence so it's smaller and easier to review then once they've all had an approval, we could combine them into one branch that doesn't introduce any breaking behaviour. How would you feel about that?

pezholio:
Yeah, good point. How about I make a branch called `organisation-versioning` and then set the target branch to that? Then I can open PRs to go onto that branch, then when the feature is done, we PR that branch into staging?

Gweaton:
> Yeah, good point. How about I make a branch called `organisation-versioning` and then set the target branch to that? Then I can open PRs to go onto that branch, then when the feature is done, we PR that branch into staging?

Good shout - will approve this now.

Gweaton (approved):

PR #166, opened 2022-01-26
(2037) Add ProfessionVersion table and service

Gweaton:
# Changes in this PR

- Adds `ProfessionVersion` entity, linked to `Profession`s, `Legislation`s and `Organisation`s. This is currently not used anywhere, but will replace the user-entered values on the `Profession` entity so we can keep track of different versions.
- Adds `ProfessionVersion` factory class, and updates existing factory classes to make room for the new entity.
- Adds a basic `ProfessionVersionsService` for finding and saving `ProfessionVersions`


pezholio (approved):
LGTM 👍

PR #167, opened 2022-01-27
Miscellaneous Profession refactoring

Gweaton:
# Changes in this PR

This mostly refactors the tests for controllers used in the "Add/edit a profession" journey to make them more in line with how we're currently doing things (not setting up test context in a `beforeEach`, not worrying about the returned values from the `i18nService` and preferring `mockResolvedValue` over `mockImplementation` where relevant. This paves the way for new behaviour to be added now we're adding versioning to this.

I've refactored the `ProfessionsService` tests here too, as we'll be adding some more functionality to this too.

This also refactors the "edit" journey to use an `id` instead of the `slug`, as that's what we're doing for the Organisations journey and it feels sensible to be consistent here.

This PR also tweaks the backLink decorator to handle upper case letters, as we'll be adding multiple `id`s to URLs in the version journey (namely `professionId` and `versionId`).

Opening these changes in their own PR to keep the main one from being a long read and a very time-consuming effort to review.

### DIFF
--- a/src/common/back-link.interceptor.spec.ts
+++ b/src/common/back-link.interceptor.spec.ts
@@ -36,14 +36,14 @@ describe('BackLinkInterceptor', () => {
       beforeEach(() => {
         request = createMock<Request>({
           params: {
-            baz: 'boo',
+            bazName: 'boo',
             id: '123',
           },
         });
       });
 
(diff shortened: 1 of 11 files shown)


pezholio (approved):
Lovely stuff! ❤️

PR #178, opened 2022-01-28
(2039) Create a version when editing an Organisation

pezholio:
❗ This PR will merge into the `organisation-version` branch, as some of the functionality is still broken. This will be rectified in the next PR ❗ 

This adds a confirmation page when clicking through to edit an organisation. If the user clicks through, this then creates a new OrganisationVersion, which is a copy of the latest draft or live version, and then sends the user through to the edit page. 

# Screenshots

image

image

image

image

### DIFF
--- a/src/organisations/admin/organisation-versions.controller.ts
+++ b/src/organisations/admin/organisation-versions.controller.ts
@@ -37,4 +37,30 @@ export class OrganisationVersionsController {
 
     return organisation;
   }
+
+  @Post('/:organisationID/versions')
+  async create(
+    @Res() res: Response,
+    @Req() req: RequestWithAppSession,
+    @Param('organisationID') organisationID: string,
+  ): Promise<void> {
+    const latestVersion =
+      await this.organisationsVersionsService.findLatestForOrganisationId(
+        organisationID,
+      );
+
+    const newVersion = {
+      ...latestVersion,
+      id: undefined,
+      user: req.appSession.user,
+      created_at: undefined,
+      updated_at: undefined,
+    } as OrganisationVersion;
+
+    const version = await this.organisationsVersionsService.save(newVersion);
+
+    return res.redirect(
+      `/admin/organisations/${version.organisation.id}/versions/${version.id}/edit`,
+    );
+  }
 }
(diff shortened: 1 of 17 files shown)


views/admin/organisations/edit.njk
```diff
@@ -18,17 +18,19 @@
...
-            label: {
-              text: ("organisations.admin.form.label.name" | t),
-              classes: "govuk-label--m"
-            },
-            id: "name",
-            name: "name",
-            value: name
-          })
-        }}
+        {% if slug|length == 0 %}
```
Gweaton, line 25: Do you think it could be worth checking the status, whether it's `unconfirmed`, rather than checking slug length? I wonder if that'd be clearer to readers.

src/organisations/admin/organisation-versions.controller.ts
```diff
@@ -37,4 +37,28 @@ export class OrganisationVersionsController {
...
+    @Req() req: RequestWithAppSession,
+    @Param('organisationID') organisationID: string,
+  ): Promise<void> {
+    const latestVersion =
+      await this.organisationsVersionsService.findLatestForOrganisationId(
+        organisationID,
+      );
+
+    const newVersion = {
+      ...latestVersion,
```
Gweaton, line 53: Do we need to overwrite the `updated_at` here too? And probably `created_at`.

views/admin/organisations/edit.njk
```diff
@@ -7,7 +7,11 @@
 
 {% block content %}
   <h1 class="govuk-heading-xl">
-    {{ "organisations.admin.edit.heading" | t }}
+    {% if slug|length == 0 %}
```
Gweaton, line 10: Again, I wonder if we can use the `unconfirmed` value here, or are we not passing versions in here?

Gweaton (approved):
This looks good, just a couple of minor things

src/organisations/admin/organisation-versions.controller.ts
pezholio: Good call - I just noticed that when working on the next bit

views/admin/organisations/edit.njk
pezholio: Yeah, that's right 👍

PR #184, opened 2022-01-31
(2044) Create versions when adding or editing a Profession

Gweaton:
# Changes in this PR

- Updates relationship between `ProfessionVersion` and `Qualification` to be OneToOne based on new designs.
- Creates `ProfessionVersion` objects with seed data.
- Creates a new `ProfessionVersion` when adding a new profession or copies the latest existing version when editing a profession.
- Uses presence of slug, rather than `confirmed` field on Profession to determine whether we're editing or adding a profession.
- Marks a `version` as `draft` when the add/edit journey is completed.

## Still to do:
- Hide "name" field when editing a Profession which already has a slug (and is "confirmed"), as we don't want to allow users to edit names, as may be inconsistent with slug.
- Store User who created the version on the version (awaiting changes from the Organisation version PR)
- Refactor `slug` construction behaviour to use new helper function introduced in Organisation version PR.
- Fetch latest version and only display live versions on public site and draft/live versions in admin site.


pezholio (approved):
LGTM - thanks for doing this. The professions journey is deffo more arduous! Just a couple of things 👍

PR #198, opened 2022-02-02
(2048) Show latest draft or live version on admin side

pezholio:
This changes the behaviour of the admin side to fetch the latest version of an organisation (either draft or live) on the listing page, and then links through to show the version, rather than the organisation. This should now have fixed all the failing e2e tests we had skipped, so once this is in, we should be in a position to merge the `organisation-versioning` branch to main 🙌


src/organisations/organisation-versions.service.ts
```diff
@@ -51,6 +51,20 @@ export class OrganisationVersionsService {
     );
   }
 
+  async allDraftOrLive(): Promise<Organisation[]> {
```
bryangaledxw, line 65: Can this end up returning archived/unconfirmed versions?

src/organisations/presenters/organisation.presenter.ts
```diff
@@ -27,6 +27,11 @@ export class OrganisationPresenter {
       {
         html: await this.industries(),
       },
+      {
+        html: await this.i18nService.translate(
```
bryangaledxw, line 31: If an org has a live version and a later draft version, is the status definitely draft? I feel like a user would see "draft" to mean "not live"

src/organisations/organisation-versions.service.spec.ts
```diff
@@ -253,3 +259,43 @@ describe('OrganisationVersionsService', () => {
     });
   });
 });
+
+expect.extend({
```
bryangaledxw, line 263: Will this be useful outside this test? Is there a more global place this could live?

bryangaledxw (approved):
Looks good, but there's a couple of things I just want to check

src/organisations/organisation-versions.service.ts
pezholio: Oh, good call. I missed that!

src/organisations/presenters/organisation.presenter.ts
pezholio: This is referring to the version that's visible in the admin interface, but I get where you're coming from. Let's see what comes out in testing 👍

src/organisations/organisation-versions.service.spec.ts
pezholio: Good call - I've added another commit for this :+1

PR #199, opened 2022-02-03
Organisation versioning

pezholio:
❗ Not to be merged until #203 and #204 are merged ❗ 

This merges in all the changes from #198 #185 #178 and #160 and adds a changelog entry.

There's still some work to be done, namely some design tweaks and the publication process, but this is enough to ensure no functionality is broken and ensure this branch doesn't drift too far away from the mainline.

All the changes in here have been reviewed, so this won't need a thorough review.


Gweaton (approved):
🏁

PR #203, opened 2022-02-03
(2042) Fetch Organisations with a Version on Profession Pages

pezholio:
This fixes the remaining failing tests in #199 to make sure we use versions when fetching organisations on professions pages. Hopefully this won't get in your way too much @Gweaton!


Gweaton (approved):
Looks good to me

PR #204, opened 2022-02-03
(2005) Add flash message when creating a new OrganisationVersion

pezholio:
This updates the Organisation flow to be more in line with the designs, so we see a flash message when creating/updating versions, rather than a confirmation page. This also makes flash messages much easier to use across the application.

## Screenshots

image

image


src/common/flash-message.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+
+      expect(response).toMatch(
+        '<h3 class="govuk-notification-banner__heading">Title</h3>',
+      );
+
+      expect(response).toMatch('<p class="govuk-body">Body</p>');
+    });
+  });
+
+  describe('when the title id provided', () => {
```
Gweaton, line 16: tiny typo on `is` here

Gweaton (approved):
Nice

PR #205, opened 2022-02-03
(2045) Only show Live Professions on public-facing pages

Gweaton:
# Changes in this PR

- Adds a new seeded profession so we have more to play with locally.
- Only displays Professions with a live ProfessionVersion in the list of Professions on the public-facing site.
- Updates the "show" profession page to support displaying more than one Legislation (even though we only capture one at the moment). This will be iterated on in future designs being worked on at the moment.

### DIFF
--- a/src/professions/profession-versions.service.spec.ts
+++ b/src/professions/profession-versions.service.spec.ts
@@ -143,26 +143,13 @@ describe('ProfessionVersionsService', () => {
 
       expect(result).toEqual(expectedProfessions);
 
-      expect(queryBuilder.leftJoinAndSelect).toHaveBeenCalledWith(
+      expect(queryBuilder).toHaveJoined([
         'professionVersion.profession',
-        'profession',
-      );
-      expect(queryBuilder.leftJoinAndSelect).toHaveBeenCalledWith(
         'professionVersion.industries',
-        'industries',
-      );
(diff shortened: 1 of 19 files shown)


src/professions/profession-versions.service.spec.ts
```diff
@@ -120,4 +121,55 @@ describe('ProfessionVersionsService', () => {
...
+        'organisation',
+      );
+      expect(queryBuilder.leftJoinAndSelect).toHaveBeenCalledWith(
+        'professionVersion.qualification',
+        'qualification',
+      );
+      expect(queryBuilder.leftJoinAndSelect).toHaveBeenCalledWith(
+        'professionVersion.legislations',
+        'legislations',
+      );
```
pezholio, line 195: We can probably use [`toHaveJoined`](https://github.com/UKGovernmentBEIS/regulated-professions-register/blob/b6df192d557cfe347c23d4f3581fd7da80df2c03/src/testutils/jest.setup.ts#L2) here

pezholio (approved):
LGTM :shipit:

src/professions/profession-versions.service.spec.ts
pezholio: Oh, ignore me. I see you've used it in a later commit 👍

PR #207, opened 2022-02-04
(2049) Only show draft & live Professions to admins

Gweaton:
# Changes in this PR

Displays draft and live professions to admins on the index page, updating the Profession entity to use its latest version's `status` field, which is now displayed in the table.

## Screenshots of UI changes

image

### DIFF
--- a/src/organisations/organisation.entity.ts
+++ b/src/organisations/organisation.entity.ts
@@ -75,6 +75,10 @@ export class Organisation {
     organisation: Organisation,
     organisationVersion: OrganisationVersion,
   ): Organisation {
+    const professions = (organisation.professions || [])
+      .map((profession) => Profession.withLatestLiveOrDraftVersion(profession))
+      .filter(Boolean);
+
     return {
       ...organisation,
       alternateName: organisationVersion.alternateName,
@@ -86,6 +90,7 @@ export class Organisation {
       fax: organisationVersion.fax,
       versionId: organisationVersion.id,
       status: organisationVersion.status,
+      professions: professions,
     } as Organisation;
   }
 
--- a/src/professions/profession-versions.service.ts
+++ b/src/professions/profession-versions.service.ts
@@ -60,6 +60,23 @@ export class ProfessionVersionsService {
     );
   }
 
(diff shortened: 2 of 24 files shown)


Gweaton:
This needs a bit more work to fix up the admin/show page, which I forgot to update. Will do that asap.

pezholio (approved):
Happy for this to be merge once #2045 is in 👍

src/professions/profession-versions.service.ts
```diff
@@ -89,6 +89,18 @@ export class ProfessionVersionsService {
...
+  async findByIdWithProfession(
+    professionId: string,
+    id: string,
+  ): Promise<Profession> {
+    const version = await this.versionsWithJoins()
+      .leftJoinAndSelect('organisation.versions', 'organisationVersions')
+      .where({ profession: { id: professionId }, id })
+      .getOne();
+
+    return Profession.withVersion(version.profession, version);
```
pezholio, line 101: I actually tweaked the organisation version of this in my recent PR to return to version, and then used `.withVersion` in the controller because a) it made more sense to return a version in a version service, and b) I could use this in the publication flow too.

Gweaton: I might refactor this in the next step, if that's ok? Just to get it out as it's a blocker for some of the other PRs.

pezholio: Yeah, no drama 👍

src/organisations/organisation.entity.ts
```diff
@@ -75,6 +75,10 @@ export class Organisation {
     organisation: Organisation,
     organisationVersion: OrganisationVersion,
   ): Organisation {
+    const professions = (organisation.professions || [])
+      .map((profession) => Profession.withLatestLiveOrDraftVersion(profession))
+      .filter(Boolean);
```
pezholio, line 80: This is the first time I've seen `filter(Boolean)`. Neat trick!

pezholio (approved):
LGTM assuming the build gets fixed (which I think is down to an undefined variable sneaking in)

PR #209, opened 2022-02-07
(2041) Allow a draft organisation to be published

pezholio:
This adds a new "Publish this organisation" button to publish a draft version of an Organisation. The marks the latest live version as archived and then promotes the draft version to live.

## Screenshots

image

image


bryangaledxw (approved):
Looks good. Noting that there's no checking of that the user has permissions to publish, though I think we need a complete sweep through the app of what permissions are used where anyway

pezholio:
Yeah, that did occur to me. Best to handle that once your permissions work is nailed down I reckon 👍

PR #228, opened 2022-02-09
Fix broken user edit journey

pezholio:
We haven't done much to the user edit journey, and there were a bunch of broken things when we tried to edit a user, namely the fact that we tried to recreate the user in Auth0 when editing an existing user. There were also a bunch of issues with the headings, which I've fixed now as well.

# Screenshots

image

image

image

image

image


src/users/users.controller.ts
```diff
@@ -162,4 +156,26 @@ export class UsersController {
 
     await this.usersService.delete(id);
   }
+
+  async createUserInAuth0(user: User) {
```
bryangaledxw, line 159: Can this be `async createUserInAuth0(user: User): void`?

src/users/users.controller.spec.ts
```diff
@@ -229,28 +229,54 @@ describe('UsersController', () => {
...
+    it('should not attempt to create a user in auth0 if it already has an externalIdentifier', async () => {
+      const user = userFactory.build({ externalIdentifier: 'extid|1234567' });
+      const res = createMock<Response>();
+
+      usersService.find.mockResolvedValue(user);
+
+      await controller.complete(res, user.id);
+
+      expect(auth0Service.createUser).not.toBeCalled();
+      expect(userMailer.confirmation).not.toBeCalledWith(
```
bryangaledxw: Can this be just `expect(userMailer.confirmation).not.toBeCalled()`?

src/users/helpers/get-action-type-from-user.ts
```diff
@@ -0,0 +1,7 @@
+import { User } from '../user.entity';
+
+export type ActionType = 'edit' | 'new';
+
+export function getActionTypeFromUser(user: User): ActionType {
+  return user.externalIdentifier ? 'edit' : 'new';
```
bryangaledxw, line 6: Does checking `confirmed` on the user work? That would seem less roundabout

views/admin/users/complete.njk
```diff
@@ -6,11 +6,13 @@
...
   <div class="govuk-grid-column-two-thirds">
     {{ govukPanel({
-      titleText: ("users.headings.done" | t),
-      html: "users.body.panel" | t({email: (email | escape)})
+      titleText: (("users.headings." + action + ".done") | t),
+      html: ("users.body." + action + ".panel") | t({email: (email | escape)})
     }) }}
 
-    <p class="govuk-body">{{ "users.body.confirmation" | t  }}</p>
+    {% if action == "new" %}
```
bryangaledxw, line 13: Would prefer `===`

cypress/integration/admin/user/edit.ts
```diff
@@ -0,0 +1,55 @@
...
+      cy.loginAuth0();
+    });
+
+    it('I can edit a user', () => {
+      cy.visitAndCheckAccessibility('/admin/users');
+      cy.contains('View Editor').click();
+
+      cy.checkAccessibility();
+
+      cy.contains('Change Organisation').click();
```
bryangaledxw, line 20: Do we want `cy.checkAccessibility();` after this?

bryangaledxw (approved):
Looks good. Few minor picky things - the one I care most about is using `confirmed` rather than `externalIdentifier`

src/users/users.controller.ts
pezholio: Good catch 👍

src/users/helpers/get-action-type-from-user.ts
pezholio: Yeah, good idea 👍

PR #230, opened 2022-02-09
Feature: Profession versioning

Gweaton:
# Changes in this PR

This is a collection of all previous work to support versions of a Profession and the publishing of these draft versions.

All PRs had been approved before opening this PR, but we used this feature branch to keep things easy to review and so as not to break `main`.

All tests are passing here, and I've done some manual testing after a long rebase, so I'm fairly confident this is safe to go in.

One known bug:

- Adding a new profession doesn't seem to be saving it on the organisation entity (or it's not being fetched) - it appears in the live list of Professions, but doesn't appear on the Organisation on either the admin or public pages. I'll investigate this in the meantime.


pezholio (approved):

pezholio:
image

Seriously though - thanks for doing all this. Let's get it on the mainline ASAP 👍

PR #232, opened 2022-02-09
Revise permissions required for editing professions

bryangaledxw:
Add a new `EditProfession` permission, granted to all roles, and require this when editing professions. Only require the `CreateProfession` permission when actually creating a profession.

Note this is not the full sweep of professions is the app, so much as the minimum necessary to prevent confusion during user research


pezholio (approved):
LGTM - can we add a Changelog entry pls? I'm conscious we're (and I'm the worst culprit!) getting a bit sloppy on these 👍

PR #237, opened 2022-02-10
(2105) Add protected titles and regulations URL to professions

bryangaledxw:
# Changes in this PR

This update the "Regulations" screen of profession editing to match this design - https://miro.com/app/board/uXjVOaZc0VI=/?moveToWidget=3458764518194509520&cot=14

Note that this PR does not update the check your answers page, or the profession display. This will come in a PR to follow shortly. This also keeps protected titles as a `textarea`, rather than distinct lines that can be added to as shown in the design

### Before

localhost_3000_admin_professions_751b9101-6c12-483f-9756-4ef23204258a_versions_8b757860-2e07-4888-bd0b-f5824752e92d_regulated-activities_edit_change=true (1)

### After

localhost_3000_admin_professions_751b9101-6c12-483f-9756-4ef23204258a_versions_8b757860-2e07-4888-bd0b-f5824752e92d_regulated-activities_edit_change=true


src/professions/professions.seeder.ts
```diff
@@ -114,6 +118,8 @@ export class ProfessionsSeeder implements Seeder {
           industries,
           qualification,
           profession.reservedActivities,
+          profession.protectedTitles,
```
Gweaton, line 121: Do we have these in the seed data, or are they `undefined` here? Is it worth adding to the seeds if not?

Gweaton (approved):
Looks good to me, just one comment about seeds.

src/professions/professions.seeder.ts
bryangaledxw: Thanks for flagging, turns out I hadn't got the seeder changes quite right. Looking at the real spreadsheet, the only data I can obviously add is a protected title for the Trademark Attorney, so have added that

PR #239, opened 2022-02-10
(2109) Refactor Profession versioning

Gweaton:
# Changes in this PR

- Remove "confirmed" field from Professions, now we can rely on the presence of a `slug` for this.
- Use shared `slugHelper` function instead of `confirm` method in `ConfirmationController`.
- Remove all duplicate profession fields that are now stored on the version (and from seed functions)
- Fix seeder deletion order
- Move `ProfessionVersionsController` into admin directory


Gweaton:
@bryangaledxw Thanks for the feedback, have added another commit with an `isConfirmed` helper.

I'm going to leave `setSlug` as-is, as it's the same across both the `ProfessionsService` and `OrganisationsService`, but we can possibly refactor both in future.

PR #242, opened 2022-02-10
(2103) Simplify nation selection

pezholio:
This updates the nation selection part of the Profession edit journey to allow users to select United Kingdom if a profession is regulated across the UK, or "Certain Nations" if the profession is only regulated in particular nation(s)

# Screenshots

image

image

### DIFF
--- a/src/professions/admin/top-level-information.controller.ts
+++ b/src/professions/admin/top-level-information.controller.ts
@@ -62,10 +62,14 @@ export class TopLevelInformationController {
       versionId,
     );
 
+    const coversUK = version.occupationLocations
+      ? this.isUK(version.occupationLocations)
+      : null;
+
     return this.renderForm(
       res,
       profession.name,
-      null,
+      coversUK,
       version.industries || [],
       version.occupationLocations || [],
       isConfirmed(profession),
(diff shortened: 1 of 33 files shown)


src/professions/admin/top-level-information.controller.ts
```diff
@@ -149,6 +151,7 @@ export class TopLevelInformationController {
   private async renderForm(
     res: Response,
     name: string,
+    coversUK: string,
```
Gweaton, line 164: Does it make sense to have this as boolean here? We can convert it like:`Boolean(Number(submittedValues.coversUK))`

Gweaton (approved):
This looks good - just had one comment on the use of `1` and `0` for our values here. It's hard to convert checkboxes to booleans, but I feel it's a bit clearer to pass round boolean value as much as possible where we can.

src/professions/admin/top-level-information.controller.ts
Gweaton: There are other places in this PR that use 1 and 0, can we replace as many of those as possible with a true or false value?

pezholio: Yeah, I tried that but, as you say, it was tricky to convert them (I expected the DTO to do it, but it wasn't having any of it!). I'll have a crack with your suggestion.

pezholio: That's got it - much better!

PR #243, opened 2022-02-10
Fix incorrect role descriptions being shown when creating a non-service owner user

bryangaledxw:
Previously the same description for each role was shown regardless of whether we were creating a service owner user or a regulatory body user. We now show slightly differing descriptions to reflect the different permission that a role has dependant on whether or not the user is a service owner

### DIFF
--- a/cypress/integration/admin/user/new.ts
+++ b/cypress/integration/admin/user/new.ts
@@ -56,7 +56,7 @@ describe('Creating a new user', () => {
       cy.checkAccessibility();
 
       cy.translate(
-        'users.roleDescriptions.managerProfessionsAndOrganisations',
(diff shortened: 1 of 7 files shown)


cypress/integration/admin/user/new.ts
```diff
@@ -56,7 +56,7 @@ describe('Creating a new user', () => {
       cy.checkAccessibility();
 
       cy.translate(
-        'users.roleDescriptions.managerProfessionsAndOrganisations',
```
pezholio: Can this commit be squashed into the previous one?

pezholio (approved):
LGTM - barring one comment to keep the history tidy ([Git Absorb](https://github.com/tummychow/git-absorb) is GREAT for this kind of stuff BTW), and a Changelog entry 👍

PR #249, opened 2022-02-11
(2104) Allow second regulator to be chosen

pezholio:
This adds an optional `additionalOrganisation` field to the Profession entity to allow a secondary organisation to be applied to an organisation.

# Screenshots

image

image

image

### DIFF
--- a/src/professions/admin/regulated-authorities-select-presenter.ts
+++ b/src/professions/admin/regulated-authorities-select-presenter.ts
@@ -8,12 +8,24 @@ export class RegulatedAuthoritiesSelectPresenter {
   ) {}
 
   selectArgs(): SelectItemArgs[] {
-    return this.allOrganisations.map((organisation) => ({
-      text: organisation.name,
-      value: organisation.id,
-      selected: this.selectedOrganisation
-        ? this.selectedOrganisation.id === organisation.id
-        : false,
-    }));
+    const options = [
+      {
+        text: '--- Please Select ---',
+        value: '',
+        selected: null,
+      },
+    ];
+
+    this.allOrganisations.forEach((organisation) =>
+      options.push({
+        text: organisation.name,
+        value: organisation.id,
+        selected: this.selectedOrganisation
+          ? this.selectedOrganisation.id === organisation.id
+          : false,
+      }),
+    );
+
+    return options;
   }
 }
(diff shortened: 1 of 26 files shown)


src/professions/admin/regulated-authorities-select-presenter.ts
```diff
@@ -8,12 +8,24 @@ export class RegulatedAuthoritiesSelectPresenter {
...
-    return this.allOrganisations.map((organisation) => ({
-      text: organisation.name,
-      value: organisation.id,
-      selected: this.selectedOrganisation
-        ? this.selectedOrganisation.id === organisation.id
-        : false,
-    }));
+    const options = [
+      {
+        text: '--- Please Select ---',
```
Gweaton, line 13: Nice, was trying to figure out how to do this with govuk components, turns out it's not as complicated as I thought!

src/professions/admin/regulatory-body.controller.spec.ts
```diff
@@ -147,14 +167,23 @@ describe(RegulatoryBodyController, () => {
...
         };
 
         const newOrganisation = organisationFactory.build({
           name: 'Council of Gas Safe Engineers',
         });
 
-        organisationsService.find.mockResolvedValue(newOrganisation);
+        const newAdditionalOrganisation = organisationFactory.build();
+
+        when(organisationsService.find)
```
Gweaton, line 180: I didn't know about `jest-when`, nice find.

views/admin/professions/regulatory-body.njk
```diff
@@ -51,6 +51,19 @@
...
         }}
 
+        {{
+          govukSelect({
+            label: {
+              text: ("professions.form.label.regulatoryBody.additionalAuthority" | t),
+              classes: "govuk-label--m",
+              isPageHeading: false
+            },
+            id: "additionalRegulatoryBody",
```
Gweaton, line 61: This isn't for now, but it'd be great to sweep the codebase and content here and have a consistent term for this, as we currently have Organisation, Regulatory Authority and Regulatory Body. I know Organisation is more of a data layer, but it'd at least be good to have one of RA or RB.

Gweaton (approved):
This is great - looks like the build is failing but once that's sorted this is good to go. Should we leave it until after user research on Monday just to be safe?

PR #251, opened 2022-02-11
(2077) Display user who last edited a profession

Gweaton:
# Changes in this PR

- Fixes incorrect ProfessionEntity -> User relationship.
- Adds the user who last edited the profession to the admin profession page. This will be empty if data hasn't been edited - we may want to iterate this or hide the field entirely, but this is a first step.
- Publishing a profession now creates a duplicate version, setting the user who published on it rather than the user who edited the Profession version.
- Now shows "Changed by" for BEIS editor users, which was previously hidden in the Profession index table. 

### Admin "show" page (organisation admin, after manually updating):

image

### Admin "show" page - editor user, showing entry that has not yet been updated:

image

### Admin index page (organisation admin, after manually updating):

image

### Admin index page (editor user, showing entry that has not yet been updated):

image

### DIFF
--- a/cypress/integration/admin/professions/edit.spec.ts
+++ b/cypress/integration/admin/professions/edit.spec.ts
@@ -1,3 +1,5 @@
+import { format } from 'date-fns';
+
 describe('Editing an existing profession', () => {
   context('when I am not logged in', () => {
     it('I am prompted to log in', () => {
@@ -221,6 +223,11 @@ describe('Editing an existing profession', () => {
 
       cy.checkAccessibility();
       cy.get('[data-cy=changed-by-user]').should('contain', 'Editor');
+      cy.get('[data-cy=last-modified]').should(
+        'contain',
+        format(new Date(), 'dd-MM-yyyy'),
+      );
+
       cy.translate('professions.admin.update.confirmation.heading').then(
         (flashHeading) => {
           cy.translate('professions.admin.update.confirmation.body', {
--- a/cypress/integration/admin/professions/edit.spec.ts
+++ b/cypress/integration/admin/professions/edit.spec.ts
@@ -21,6 +21,12 @@ describe('Editing an existing profession', () => {
         });
 
       cy.checkAccessibility();
+
+      cy.translate('professions.admin.changed.by').then((lastUpdatedText) => {
+        cy.get('[data-cy=changed-by-text]').should('contain', lastUpdatedText);
+      });
+      cy.get('[data-cy=changed-by-user]').should('contain', '');
+
       cy.translate('professions.admin.editProfession').then((buttonText) => {
         cy.contains(buttonText).click();
       });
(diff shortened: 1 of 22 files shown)


bryangaledxw:
Appreciate this is only draft so maybe you already have in mind - would be good to get rid of "Placeholder name" in the org-level profession listing as part of this

Gweaton:
> Appreciate this is only draft so maybe you already have this in mind - would be good to get rid of "Placeholder name" in the org-level profession listing as part of this

Thanks for pointing this out, have fixed this.

src/professions/profession-version.entity.ts
```diff
@@ -38,7 +38,7 @@ export class ProfessionVersion {
   @ManyToOne(() => Profession, (profession) => profession.versions)
   profession: Profession;
 
-  @ManyToOne(() => User, (user) => user.organisationVersions)
+  @ManyToOne(() => User, (user) => user.professionVersions)
```
pezholio, line 41: I think the reason it's not generating a migration is because this is all in code, the indexes are all set up against the tables directly. The second argument is just telling TypeOrm where to load the relations once they've been fetched from the database.

cypress/integration/admin/professions/edit.spec.ts
```diff
@@ -21,6 +21,12 @@ describe('Editing an existing profession', () => {
         });
 
       cy.checkAccessibility();
+
+      cy.translate('professions.admin.changed.by').then((lastUpdatedText) => {
+        cy.get('[data-cy=changed-by-text]').should('contain', lastUpdatedText);
```
pezholio, line 28: Big 👍  for using data attributes here

src/professions/profession-versions.service.ts
```diff
@@ -33,6 +36,40 @@ export class ProfessionVersionsService {
     return this.repository.save(version);
   }
 
+  async createNewVersion(
```
pezholio, line 39: Yes! Big fan of keeping logic out of the controller 👍 

Could we just call it `create`, as we know we'll be creating a version?

cypress/integration/admin/professions/edit.spec.ts
```diff
@@ -187,6 +189,11 @@ describe('Editing an existing profession', () => {
 
       cy.checkAccessibility();
       cy.get('[data-cy=changed-by-user]').should('contain', 'Editor');
+      cy.get('[data-cy=last-modified]').should(
+        'contain',
+        format(new Date(), 'dd-MM-yyyy'),
```
pezholio, line 194: I think this is fine. It's a shame we can't stub or fake, but it's covered by the unit tests, so I think we're safe 👍

pezholio (approved):
LGTM - just one teensy thing re: naming, but happy for you to ignore me 😄

src/professions/profession-versions.service.ts
Gweaton: Yep, sounds good to me - have renamed.

pezholio:
Just noticed the Changelog entries are in the wrong place here, they should be under the `Unreleased` header

pezholio (approved):
LGTM 👍

PR #253, opened 2022-02-14
(2107) Add registration page

pezholio:
This adds a new "Registrations" page to the add a profession journey to collect information about registration requirements, and a link to more information about the requirements.

## Screenshots

image

image


src/professions/admin/registration.controller.ts
```diff
@@ -0,0 +1,163 @@
...
+@Controller('admin/professions')
+export class RegistrationController {
+  constructor(
+    private readonly professionsService: ProfessionsService,
+    private readonly professionVersionsService: ProfessionVersionsService,
+    private readonly i18nService: I18nService,
+  ) {}
+
+  @Get('/:professionId/versions/:versionId/registration/edit')
+  @Permissions(UserPermission.EditProfession)
```
bryangaledxw, line 48: Think this should be `@Permissions(UserPermission.CreateProfession, UserPermission.EditProfession)`, as we'll come through here for both creating a new profession and editing an existing one

src/professions/admin/registration.controller.ts
```diff
@@ -0,0 +1,163 @@
...
+      res,
+      selectedMandatoryRegistration,
+      isConfirmed(profession),
+      change,
+    );
+  }
+
+  @Post('/:professionId/versions/:versionId/registration')
+  @Permissions(UserPermission.EditProfession)
+  @BackLink((request: Request) =>
```
bryangaledxw, line 73: Ditto above

src/professions/admin/interfaces/registration.template.ts
```diff
@@ -0,0 +1,7 @@
+import { RadioButtonArgs } from '../../../common/interfaces/radio-button-args.interface';
+export interface RegistrationTemplate {
+  mandatoryRegistrationRadioButtonArgs: RadioButtonArgs[];
+  change: boolean;
```
bryangaledxw, line 4: I know we have this issue basically everywhere, but this actually comes through as a string, right? Same with `@Query('change') change: boolean` on the controller itself

bryangaledxw (approved):

PR #261, opened 2022-02-14
(2113) Ensure all @Permissions are correct for all routes in app

bryangaledxw:
This assigns the correct permissions for all related routes

This PR doesn't enforce constraints such as "a user cannot edit a profession within another organisation", which cannot be achieved with @Permissions decorators


src/professions/admin/professions.controller.ts
```diff
@@ -88,15 +88,15 @@ export class ProfessionsController {
     const allProfessions =
       await this.professionVersionsService.allDraftOrLive();
 
-    const user = request['appSession'].user as User;
+    const actingUser = request.appSession.user as User;
 
-    const showAllOrgs = user.serviceOwner;
+    const showAllOrgs = actingUser.serviceOwner;
```
pezholio, line 99: `actingUser` is much better naming 👍

cypress/integration/admin/organisations/edit.spec.ts
```diff
@@ -1,7 +1,7 @@
 describe('Editing organisations', () => {
-  context('When I am logged in as admin', () => {
```
pezholio, line 2: This is a really good approach 👍

src/organisations/admin/organisations.controller.ts
```diff
@@ -55,14 +56,28 @@ export class OrganisationsController {
   @Permissions(UserPermission.CreateOrganisation, UserPermission.EditOrganisation, UserPermission.DeleteOrganisation, UserPermission.PublishOrganisation)
   @Render('admin/organisations/index')
   @BackLink('/admin')
-  async index(@Query() query: FilterDto = null): Promise<IndexTemplate> {
+  async index(
```
pezholio, line 64: We should probably redirect non service owner users to their own organisation page, but this is fine for now

pezholio (approved):
LGTM - I do wonder if there's a way we can refactor the `Permissions` decorator to be less repetitive, but that's just an idle thought really. Thanks for doing this 👍

PR #268, opened 2022-02-15
Restore editor user

pezholio:
The beis editor user has been deleted again (probably because we've (I've) been running the e2e tests against the development environment). I've recreated the user with the same username and password, and have put a new externalIdentifier in.

I've also tweaked the deletion test so we create a user before deleting it, so we don't get the issue going forward.


Gweaton (approved):

bryangaledxw (approved):
Suggest do `const createUser = (name: string, email: string)` as ` function createUser(name: string, email: string): void` for consistency, but otherwise looks good

PR #275, opened 2022-02-16
(2065) Format line breaks in data presented to user

bryangaledxw:
# Changes in this PR

Any data that the user enters in a multi-line text field, will now be presented back as the user would naturally expect, with new lines preserved.

### Before
localhost_3000_admin_professions_863278ef-ca98-466b-8445-a56f536b1538_versions_5d333f32-a15a-4972-92b3-45022678ff9e

### After
localhost_3000_admin_professions_5817a78b-263b-4fd2-ba8b-6aac022ddd84_versions_ef2ff443-c027-4b4a-8e6d-363a454c2150

### DIFF
--- a/seeds/development/organisations.json
+++ b/seeds/development/organisations.json
@@ -5,7 +5,7 @@
     "versions": [
       {
         "alternateName": "",
-        "address": "123 Example Street, London, EC1 1AB",
+        "address": "123 Example Street\nLondon\nEC1 1AB",
         "url": "https://www.gov.uk/guidance/early-years-qualifications-finder#overseas-qualifications",
         "email": "dfe@example.com",
         "contactUrl": "def@example.com/contact-us",
@@ -15,7 +15,7 @@
       },
       {
         "alternateName": "",
-        "address": "345 Example Street, London, EC1 1AB",
+        "address": "345 Example Street\nLondon\nEC1 1AB",
         "url": "https://www.gov.uk/guidance/early-years-qualifications-finder#overseas-qualifications",
         "email": "old@example.com",
         "contactUrl": "def@example.com/contact-us",
@@ -31,7 +31,7 @@
     "versions": [
       {
         "alternateName": "",
-        "address": "123 Fake Street, London, EC1 1AB",
+        "address": "123 Fake Street\nLondon\nEC1 1AB",
         "url": "www.corgi-gas-safety.com",
         "email": "gas@example.com",
         "contactUrl": "gas@example.com/contact-us",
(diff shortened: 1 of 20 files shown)


src/helpers/format-multiline-string.helper.ts
```diff
@@ -0,0 +1,35 @@
...
+  );
+
+  const htmlParagraphs = `<p class="${classes}">${htmlLinesByParagraph.join(
+    `</p><p class="${classes}">`,
+  )}</p>`;
+
+  return htmlParagraphs;
+}
+
+function santiseLineBreaks(text: string): string {
```
pezholio, line 27: Very thorough - I like this 👍

src/helpers/format-multiline-string.helper.ts
```diff
@@ -0,0 +1,35 @@
+import { escape } from './escape.helper';
+
+export function formatMultilineString(text: string, classes = ''): string {
```
pezholio, line 3: Should we set the class to `govuk-body` by default, as I've a feeling that's where we'll use it most

src/organisations/presenters/organisation.presenter.spec.ts
```diff
@@ -347,10 +350,12 @@ describe('OrganisationPresenter', () => {
       const presenter = new OrganisationPresenter(organisation, i18nService);
 
       expect(presenter.address()).toEqual(
-        `${escapeOf('123 Fake Street<br /> London<br /> SW1A 1AA')}`,
+        `${multilineOf('123 Fake Street, London, SW1A 1AA')}`,
```
pezholio, line 353: This is a veeeery useful test helper

seeds/development/organisations.json
```diff
@@ -5,7 +5,7 @@
     "versions": [
       {
         "alternateName": "",
-        "address": "123 Example Street, London, EC1 1AB",
+        "address": "123 Example Street\nLondon\nEC1 1AB",
```
pezholio, line 8: I do wonder if we should split the addresses into constituent parts in the data model, but that might just be due to the fact that I spent a good portion of my life steeped in address data!

pezholio (approved):
Looks awesome! :shipit:

src/helpers/format-multiline-string.helper.ts
bryangaledxw: My leaning is no, since most there's only two places in this PR I've needed to specify `govuk-body`, and I don't want to litter the HTML with class declarations that aren't need. I'm ok with pushback on this though if this is a strong preference

seeds/development/organisations.json
bryangaledxw: Rabbit hole for another PR? 😛

PR #276, opened 2022-02-16
Tidy up Dockerfile

pezholio:
This tweaks the Dockerfile so we can do caching properly. All of the layers now inherit from the base layer and we only copy what we need from the dependency layer, ensuring that we cache as much as is humanly possible.

The only things that should not be cached when running CI are the test dependencies. This should keep Docker build times down to a couple of minutes at the most, as long as the package.json or Dockerfile doesn’t change.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,10 +5,14 @@ FROM node:16 as base
 
 ENV APP_HOME /srv/app
 ENV DEPS_HOME /deps
+ENV TEST_HOME /srv/test
 
 ARG NODE_ENV
 ENV NODE_ENV ${NODE_ENV:-production}
 
+# ------------------------------------------------------------------------------
+# Dependencies
+# ------------------------------------------------------------------------------
 FROM base AS dependencies
 
 RUN mkdir -p ${DEPS_HOME}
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -63,3 +68,33 @@ RUN apt-get update && apt-get install -y --fix-missing --no-install-recommends \
   libxtst6 \
   xauth \
   xvfb
+
+RUN mkdir -p ${TEST_HOME}
+WORKDIR ${TEST_HOME}
+
+COPY --from=dependencies ${DEPS_HOME}/package.json ${TEST_HOME}/package.json
```
Gweaton, line 75: How was this being done before? Or was it not?

Gweaton (approved):
I think this makes sense, just have a few questions as I could do with scrubbing up on my Dockerfile knowledge.

Dockerfile
pezholio: The theory is that we do dependencies as high up as possible and out of the main flow, so this _should_ only be built when we don't have a cahce. But something isn't working as expected, because pushing subsequent commits when there is a cache hit doesn't build from the cache as I'd expect.

Dockerfile
```diff
@@ -63,3 +68,33 @@ RUN apt-get update && apt-get install -y --fix-missing --no-install-recommends \
...
+COPY --from=dependencies ${DEPS_HOME}/package-lock.json ${TEST_HOME}/package-lock.json
+COPY --from=dependencies ${DEPS_HOME}/node_modules ${TEST_HOME}/node_modules
+COPY --from=dependencies /root/.cache/Cypress /root/.cache/Cypress
+
+COPY --from=web ${APP_HOME}/nest-cli.json ${TEST_HOME}/nest-cli.json
```
erbridge, line 80: Could could just do `COPY --from=web ${APP_HOME} ${APP_HOME}` couldn't you?

Dockerfile
```diff
@@ -48,8 +54,12 @@ CMD ["node", "dist/main"]
 # ------------------------------------------------------------------------------
 # Test
 # ------------------------------------------------------------------------------
-FROM web AS test
+FROM base AS test
+
+RUN mkdir -p ${TEST_HOME}
+WORKDIR ${TEST_HOME}
 
+COPY --from=web ${APP_HOME} ${TEST_HOME}
```
erbridge, line 62: You want to do this after the `apt-get`.

PR #283, opened 2022-02-16
(2125) Rename "Publish now" to "Save as draft" when editing a profession

Gweaton:
# Changes in this PR

Renames Publish now button to Save as draft when editing a profession, as this is what it was really doing. We also move it into an `<aside>` at the top of the page, as users missed this button in research.

We may want to add a "Publish now" button when editing, but this can come in future if there's a need for it.

Also stops rendering a form without a submit button when not creating a new profession.

## Screenshots of UI changes

image

### DIFF
--- a/views/admin/professions/check-your-answers.njk
+++ b/views/admin/professions/check-your-answers.njk
@@ -15,7 +15,6 @@
         <h1 class="govuk-heading-l">{{ ("professions.form.headings.checkAnswers" | t) }}</h1>
       {% endif %}
 
-      <form method="post" action="confirmation">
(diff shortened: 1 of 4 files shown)


views/admin/professions/check-your-answers.njk
```diff
@@ -15,7 +15,9 @@
         <h1 class="govuk-heading-l">{{ ("professions.form.headings.checkAnswers" | t) }}</h1>
       {% endif %}
 
-      <form method="post" action="confirmation">
+      {% if not edit and not confirmed %}
+        <form method="post" action="confirmation">
```
bryangaledxw, line 19: Can this form element move to directly above the button? Then we only need one instance of `{% if not edit and not confirmed %}`

bryangaledxw (approved):

PR #284, opened 2022-02-16
(2108) Add an optional second Legislation

bryangaledxw:
# Changes in this PR

A user can now enter a second legislation name and link

### Before
localhost_3000_admin_professions_29d119c8-e9e5-476e-90a0-c77d1e3654c9_versions_e07a9ede-c6f3-425f-b855-084c036a52ae_legislation_edit_change=true (1)

### After
localhost_3000_admin_professions_29d119c8-e9e5-476e-90a0-c77d1e3654c9_versions_e07a9ede-c6f3-425f-b855-084c036a52ae_legislation_edit_change=true

### DIFF
--- a/src/professions/admin/check-your-answers.controller.ts
+++ b/src/professions/admin/check-your-answers.controller.ts
@@ -73,8 +73,8 @@ export class CheckYourAnswersController {
       : null;
 
     return {
-      professionId: professionId,
-      versionId: versionId,
+      professionId,
+      versionId,
       name: draftProfession.name,
       nations: selectedNations,
       industries: industryNames,
@@ -89,8 +89,8 @@ export class CheckYourAnswersController {
       reservedActivities: version.reservedActivities,
       protectedTitles: version.protectedTitles,
       regulationUrl: version.regulationUrl,
-      qualification: qualification,
       legislation: version.legislations[0],
+      qualification,
       confirmed: isConfirmed(draftProfession),
       captionText: ViewUtils.captionText(isConfirmed(draftProfession)),
       isUK: isUK(version.occupationLocations),
--- a/src/professions/admin/check-your-answers.controller.ts
+++ b/src/professions/admin/check-your-answers.controller.ts
@@ -38,7 +38,7 @@ export class CheckYourAnswersController {
   async show(
     @Param('professionId') professionId: string,
     @Param('versionId') versionId: string,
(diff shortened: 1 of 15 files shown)


src/professions/admin/legislation.controller.ts
```diff
@@ -88,27 +89,41 @@ export class LegislationController {
...
         isConfirmed(profession),
         submittedValues.change === 'true',
         errors,
       );
     }
 
-    version.legislations = [updatedLegislation];
+    const updatedLegislations = [updatedLegislation];
+
+    if (updatedSecondLegislation.name || updatedSecondLegislation.url) {
```
Gweaton, line 122: Just in case I haven't been understanding this properly, is only one field on the second legislation mandatory? i.e. users can enter a `name` and/**or** a `url`? Is that the same as the first Legislation?

Gweaton (approved):
This looks good to me. Have a couple of small questions.

I suppose when I introduced the `Legislations` field on the version, we could have kept it singular and added a `secondLegislation` field like we have with Organisations. We've done this in the DTO, but I think this is fine as is.

src/professions/admin/check-your-answers.controller.ts
```diff
@@ -73,8 +73,8 @@ export class CheckYourAnswersController {
       : null;
 
     return {
-      professionId: professionId,
-      versionId: versionId,
+      professionId,
```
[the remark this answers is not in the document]
bryangaledxw: Ah no - this was just me not wanting to add another instance of `foo: foo`, but wanting keep things consistent, at least within this function

src/professions/admin/legislation.controller.ts
bryangaledxw: The intent with the setup of the DTO was that if you've got one you need the other. So this line could have just as well been:

`if (updatedSecondLegislation.name)`, or `if (updatedSecondLegislation.url)`, or `if (updatedSecondLegislation.name && updatedSecondLegislation.url)`

but `if (updatedSecondLegislation.name || updatedSecondLegislation.url)` seemed least likely to break if that changes down the line

bryangaledxw:
Cheers for the review and merging in. I did spot a mention of a third(!) legislation, so I think it's best to keep `legislations` on the PV as an arbitrary length array. The DTO though I think makes sense to align with what's actually possible on the page shown to the end user

PR #290, opened 2022-02-17
(2088) Implement Plausible.io analytics

pezholio:
This uses [Plausible.io](https://plausible.io) to track visits to the service without having to store cookies on a user's machine, or implement a cookie banner. It's also considerably less bloated than Google Analytics.

As well as tracking page views, I've added a custom server side event to track unique logins to the service.

### DIFF
--- a/src/middleware/authentication.middleware.spec.ts
+++ b/src/middleware/authentication.middleware.spec.ts
@@ -6,11 +6,16 @@ import { createMock } from '@golevelup/ts-jest';
 import { AuthenticationMidleware } from './authentication.middleware';
 import { User } from '../users/user.entity';
 
+import { createPlausibleEvent } from '../common/create-plausible-event';
+
+jest.mock('../common/create-plausible-event');
+
 describe('AuthenticationMidleware', () => {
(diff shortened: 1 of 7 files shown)


src/middleware/authentication.middleware.spec.ts
```diff
@@ -6,11 +6,16 @@ import { createMock } from '@golevelup/ts-jest';
...
+import { createPlausibleEvent } from '../common/create-plausible-event';
+
+jest.mock('../common/create-plausible-event');
+
 describe('AuthenticationMidleware', () => {
   let middleware: AuthenticationMidleware;
   let user: User;
 
   beforeEach(async () => {
+    jest.clearAllMocks();
```
bryangaledxw, line 18: Can this be

```
  afterEach(() => {
    jest.resetAllMocks();
  });
```

for consistency with other tests?

(Though checking the difference between `resetAll` and `clearAll`, if we do want the latter here, I'm tempted to do a PR that just replaces all instances of the former with the latter)

bryangaledxw (approved):

PR #304, opened 2022-02-21
(2058) content iteration on the public facing side

mingoland:
# Changes in this PR
- Update page title of the search results page
- Removing the recognition data link from the start page for now
- Renaming Industry / Sector to Sectors
- Renaming URL to Website

### Before
Screenshot 2022-02-21 at 16 00 57

### After
Screenshot 2022-02-21 at 16 01 31

### Before
Screenshot 2022-02-21 at 16 27 33

### After
Screenshot 2022-02-21 at 16 27 07

### Before
Screenshot 2022-02-21 at 15 58 50

### After
Screenshot 2022-02-21 at 15 58 28

### DIFF
--- a/src/i18n/en/professions.json
+++ b/src/i18n/en/professions.json
@@ -197,7 +197,7 @@
       "name": "Name",
       "address": "Address",
       "emailAddress": "Email address",
-      "url": "URL",
+      "url": "Website",
       "phoneNumber": "Phone number"
     },
     "regulatedActivities": {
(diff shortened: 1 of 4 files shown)


src/i18n/en/professions.json
```diff
@@ -197,7 +197,7 @@
       "name": "Name",
       "address": "Address",
       "emailAddress": "Email address",
-      "url": "URL",
+      "url": "Website",
```
bryangaledxw, line 200: There's a bunch of places in the same file where we use the terminology "link". Should those change as well?

bryangaledxw (approved):

src/i18n/en/professions.json
mingoland: Good question. Let me ask Al on our channel

PR #306, opened 2022-02-21
(2149) Add additional fields to Qualification entry page, remove radio buttons and tidy up Qualification text

Gweaton:
# Changes in this PR

The first few commits of this PR is a lot of renaming from "Qualification Information" to "Qualifications" throughout the codebase, as this is a more generic name that won't go out of date as quickly if the title changes again.

After that, this PR:

- Replaces the methods to obtain qualification radio buttons with free textareas as most users were clicking "other" and entering free text anyway.
- Adds a new "More Information" url to the Qualification entity that's editable in the edit/new journey.
- Updates the text relating to Qualifications to match most up-to-date content iterations.
- Removes "Qualification Level" from the Organisation's list of Professions.
- Refactors the summary list on a Qualification into its own method on the presenter so we can test and extend its behaviour more easily.
- Removes deprecated fields.

It's a lot of changes, apologies for the large PR - I've left out a couple of things to stop this getting even more bloated (namely, being able to edit Qualification data when none exists yet, and adding additional fields to the `show` page), but those will come next.

#### Edit qualification page)

image

#### Edit qualification page

image

### Public "show" page

image


pezholio (approved):
LGTM 🚢

PR #307, opened 2022-02-22
(2132) improve the wordings for zero search results and re-arrange filters

mingoland:
# Changes in this PR
- Adding zero search results message
- Move sectors filter above nations

### Before
Screenshot 2022-02-22 at 11 06 31

### After
Screenshot 2022-02-22 at 11 12 05


Gweaton (approved):
This looks good to me - would you mind adding another commit to update the `CHANGELOG.md` with the changes here? We tend to add one called `update Changelog` at the end of a PR (or you can add each line with each commit, but that's more of a hassle!)

PR #308, opened 2022-02-22
Add cdn-route

pezholio:
Creates a cdn-route, and maps it to the app using the custom domain

This introduces 2 new tfvars:

```
custom_domain (the apex domain of where the app will be hosted)
custom_hostname (the subdomain, eg. www or staging)
```

We'll need to confirm the domain names are OK and add them as secrets before this is merged.


rjw1 (approved):
Looks fine to me

PR #317, opened 2022-02-23
Fail when coverage falls below minimum

pezholio:
This adds a coverage check to make sure as much of our code as possible is covered by unit tests. It's set to 90% at the moment, but we should aim for 95% going forward.


bryangaledxw (approved):
Able to add a commit to change `src/users/organisation/interfaces/edit-template.ts` to actually be an interface? This is having its coverage unnecessarily checked right now

pezholio:
Oh nice catch

PR #318, opened 2022-02-23
(2167) Bring back backlinks and phase banner on admin pages

Gweaton:
# Changes in this PR

Re-adds backlinks to admin pages.

These were accidentally removed when we added an admin header, as the
block overrode the `block` in the `base.njk` in the admin template. I've
extracted the phase banner and backlinks to their own partial here so we
can extend them in both `admin` and `base` templates.

This also adds a quick check in the Org "show" e2e spec to click on a back
link, just so this can be caught if it happens again in future for any
reason.

### Before

image

### After

image


pezholio (approved):
Thanks for spotting this. LGTM 👍

PR #324, opened 2022-02-23
(2152) Add confirmation step to publishing a Profession or Organisation

Gweaton:
# Changes in this PR

- Adds an "are you sure you want to publish x" screen when admins click "Publish this x" for both Professions and Organisations.
- Fixes an incorrect back link.
- Makes the end of editing an organisation take you back to the newly-updated organisation page, as it does for publishing and the profession journey.

### Publishing an organisation

image

### Publishing a profession

image

### After editing an organisation (no longer on the Organisation index page)

image


bryangaledxw (approved):
Cool stuff, just a few minor things to fix

PR #327, opened 2022-02-23
Add basic auth

pezholio:
This adds basic auth to apps that have the `BASIC_AUTH_USERNAME` and `BASIC_AUTH_PASSWORD` environment vars set - protecting sites that aren't live from prying eyes and preventing potentially incorrect information from leaking out. 

The usernames and passwords are saved as actions secrets and accessible in 1Password. This should not be merged until we've communicated the username/passwords to the client.


bryangaledxw (approved):
LGTM

PR #334, opened 2022-02-24
(2117) Admin-facing accessibility fixes

Gweaton:
# Changes in this PR

- Add hidden text to "view details"
- Fix overflowing table by making it scrollable when zoomed in
- Replace paragraph / header tags with captions above a table
- Refactor ProfessionPresenter to be constent with OrganisationPresenter.

### Before (professions table zoomed in and overflowing)

image

### After (professions table zoomed in and scrollable)

image

### After (caption replaces of `<p>` above tables on Organisations, Professions and Users pages)

image

image

### DIFF
--- a/src/organisations/admin/presenters/organisations.presenter.ts
+++ b/src/organisations/admin/presenters/organisations.presenter.ts
@@ -54,18 +54,34 @@ export class OrganisationsPresenter {
   }
 
   private async table(firstCellIsHeader = true): Promise<Table> {
+    const rows = await Promise.all(
+      this.filteredOrganisations.map(
+        async (organisation) =>
+          await new OrganisationPresenter(
+            organisation,
+            this.i18nService,
+          ).tableRow(),
+      ),
+    );
+
+    const numberOfResults = rows.length;
+
+    const caption =
+      numberOfResults === 1
+        ? await this.i18nService.translate(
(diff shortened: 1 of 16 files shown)


src/organisations/admin/presenters/organisations.presenter.ts
```diff
@@ -54,18 +54,28 @@ export class OrganisationsPresenter {
...
+            this.i18nService,
+          ).tableRow(),
+      ),
+    );
+
+    const caption = `${rows.length} ${
+      rows.length === 1
+        ? await this.i18nService.translate('organisations.admin.foundSingular')
+        : await this.i18nService.translate('organisations.admin.foundPlural')
+    }`;
```
pezholio, line 71: Again, I dunno if it'd be neater to have the number of results in the translation too, so:

```
{count} regulatory authorities found
```

src/professions/admin/professions.presenter.ts
```diff
@@ -76,4 +68,20 @@ export class ProfessionsPresenter {
       },
     };
   }
+
+  private async table(view: ProfessionsPresenterView): Promise<Table> {
```
pezholio, line 72: 💯

pezholio (approved):
LGTM - just a couple of things to think about

PR #337, opened 2022-02-25
(2091) Archive organisations

pezholio:
This adds a new controller to allow us to archive organisations. It acts in a very similar way to the publication endpoints, with a confirmation screen and an action that creates a new archived version. There was some additional work to allow us to filter by an organisation's type (and not show archived orgs by default), but given there's some work on the search coming soon, I think it's best to deprioritise this.

## Screenshots

image

image

image

### DIFF
--- a/src/i18n/en/organisations.json
+++ b/src/i18n/en/organisations.json
@@ -161,6 +161,7 @@
   },
   "status": {
     "live": "Live",
-    "draft": "Draft"
+    "draft": "Draft",
+    "archived": "Archived"
   }
 }
--- a/src/i18n/en/organisations.json
+++ b/src/i18n/en/organisations.json
@@ -69,6 +69,10 @@
     "archive": {
       "heading": "Are you sure you want to archive {organisationName}?",
       "caption": "Archiving an organisation",
+      "confirmation": {
+        "heading": "Regulatory authority archived",
+        "body": "The regulator <strong>{name}</strong> has been archived."
+      }
     },
     "tableHeading": {
       "name": "Name",
@@ -109,7 +113,7 @@
         "draft": "Edit this draft",
         "live": "Edit this regulatory authority"
       },
-      "delete": "Delete this regulatory authority",
+      "archive": "Archive this regulatory authority",
       "amend": "Amend regulatory authority",
       "publish": "Publish regulatory authority"
     }
--- a/src/i18n/en/organisations.json
+++ b/src/i18n/en/organisations.json
@@ -66,6 +66,10 @@
         "backToDashboard": "Back to the dashboard"
       }
     },
+    "archive": {
+      "heading": "Are you sure you want to archive {organisationName}?",
+      "caption": "Archiving an organisation",
+    },
     "tableHeading": {
       "name": "Name",
       "alternateName": "Alternate name",
--- a/src/organisations/admin/organisation-archive.controller.ts
+++ b/src/organisations/admin/organisation-archive.controller.ts
@@ -47,4 +47,40 @@ export class OrganisationArchiveController {
     return { organisation };
   }
 
+  @Delete(':organisationId/versions/:versionId/archive')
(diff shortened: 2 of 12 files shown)


src/organisations/organisation-versions.service.spec.ts
```diff
@@ -434,4 +434,101 @@ describe('OrganisationVersionsService', () => {
...
+
+      const draftVersion = organisationVersionFactory.build({
+        status: OrganisationVersionStatus.Draft,
+        organisation: organisation,
+      });
+
+      const saveSpy = jest.spyOn(repo, 'save');
+      const findSpy = jest
+        .spyOn(repo, 'findOne')
+        .mockImplementation(async () => liveVersion);
```
Gweaton, line 495: This could be `mockResolvedValue(liveVersion)` but not a big deal.

src/i18n/en/organisations.json
```diff
@@ -69,6 +69,10 @@
     "archive": {
       "heading": "Are you sure you want to archive {organisationName}?",
       "caption": "Archiving an organisation",
+      "confirmation": {
+        "heading": "Regulatory authority archived",
+        "body": "The new regulator <strong>{name}</strong> has been archived."
```
Gweaton, line 74: Should this be `The regulator` rather than `The new regulator`?

src/organisations/admin/organisation-archive.controller.ts
```diff
@@ -47,4 +47,40 @@ export class OrganisationArchiveController {
     return { organisation };
   }
 
+  @Delete(':organisationId/versions/:versionId/archive')
+  @Permissions(UserPermission.DeleteOrganisation)
+  async create(
```
Gweaton, line 52: I know `create` and `new` go well together, but this really feels like it should be called `delete` - can `new` and `delete` go together!?

Gweaton: Unless we read it as `create` an archived organisation, I suppose, but that feels a bit confusing!

views/admin/organisations/show.njk
```diff
@@ -28,45 +28,49 @@
...
-                  text: ("organisations.admin.button.archive" | t),
-                  href: "/admin/organisations/" + organisation.id + "/versions/" + organisation.versionId + "/archive",
-                  classes: "govuk-button--warning"
-                })
-              }}
-            </li>
-            {% endif %}
-          </ul>
-        </nav>
+        {% if organisation.status !== 'archived' %}
```
Gweaton, line 31: This is starting to feel like it could be extracted out into a sort of `adminNav` we could pass into the template that could be re-used across Professions and Organisation, as the logic in this template is getting a bit fiddly. Not one for now though.

src/organisations/organisation-versions.service.ts
```diff
@@ -79,13 +79,14 @@ export class OrganisationVersionsService {
...
 
-  async allDraftOrLive(): Promise<Organisation[]> {
+  async allWithLatestVersion(): Promise<Organisation[]> {
     const versions = await this.versionsWithJoins()
       .distinctOn(['organisationVersion.organisation'])
       .where('organisationVersion.status IN(:...status)', {
         status: [
           OrganisationVersionStatus.Live,
           OrganisationVersionStatus.Draft,
+          OrganisationVersionStatus.Archived,
```
Gweaton, line 89: Do you think we need to provide a way to re-publish versions here if we're showing them in the list? Can come later if so.

Gweaton (approved):
Looks good - sorry lots of minor comments, but nothing big!

views/admin/organisations/show.njk
pezholio: Good call, but yeah, one for later 👍

src/organisations/organisation-versions.service.ts
pezholio: I don't think it's one for now. Worth seeing if it comes out in any user research first.

PR #338, opened 2022-02-25
(2091) Archive professions

pezholio:
This adds a new controller to allow us to archive professions. It acts in a very similar way to the publication endpoints, with a confirmation screen and an action that creates a new archived version. 

# Screenshots

image
image
image
image


src/professions/profession-versions.service.spec.ts
```diff
@@ -236,6 +236,103 @@ describe('ProfessionVersionsService', () => {
...
+    });
+
+    it('archives the draft version', async () => {
+      const version = professionVersionFactory.build({
+        status: ProfessionVersionStatus.Draft,
+      });
+
+      const findSpy = jest
+        .spyOn(repo, 'findOne')
+        .mockImplementation(() => undefined);
```
Gweaton, line 257: We can remove `mockImplementation` here

src/professions/profession-versions.service.spec.ts
```diff
@@ -236,6 +236,103 @@ describe('ProfessionVersionsService', () => {
...
+
+      const draftVersion = professionVersionFactory.build({
+        status: ProfessionVersionStatus.Draft,
+        profession,
+      });
+
+      const saveSpy = jest.spyOn(repo, 'save');
+      const findSpy = jest
+        .spyOn(repo, 'findOne')
+        .mockImplementation(async () => liveVersion);
```
Gweaton, line 295: This can be `mockResolvedValue`, but doesn't matter all that much.

src/i18n/en/professions.json
```diff
@@ -249,7 +249,11 @@
     },
     "archive": {
       "heading": "Are you sure you want to archive {professionName}?",
-      "caption": "Archiving a profession"
+      "caption": "Archiving a profession",
+      "confirmation": {
+        "heading": "Regulatory authority archived",
```
Gweaton, line 254: This should be `(Regulated?) profession archived`

Gweaton (approved):
A few things that were the same as the other PR, and one copy-paste mishap! Looks good otherwise.

PR #344, opened 2022-02-28
(1766) Archive an organisation's professions

pezholio:
When we archive an Organisation, we should also archive their professions, otherwise we'll get orphaned professions showing up on the public facing side of the site, as well as a bunch of other unexpected oddness.

# Screenshots

## Before

image

## After

image

### DIFF
--- a/src/organisations/organisation-versions.service.ts
+++ b/src/organisations/organisation-versions.service.ts
@@ -9,11 +9,13 @@ import { Organisation } from './organisation.entity';
 import { User } from '../users/user.entity';
 import { ProfessionVersionStatus } from '../professions/profession-version.entity';
 
+import { ProfessionVersionsService } from '../professions/profession-versions.service';
 @Injectable()
 export class OrganisationVersionsService {
   constructor(
     @InjectRepository(OrganisationVersion)
     private repository: Repository<OrganisationVersion>,
+    private professionVersionsService: ProfessionVersionsService,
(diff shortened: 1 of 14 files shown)


src/organisations/organisation-versions.service.ts
```diff
@@ -176,6 +181,21 @@ export class OrganisationVersionsService {
...
+          await this.professionVersionsService.latestVersion(profession);
+
+        if (professionVersion) {
+          const professionVersionToBeArchived =
+            await this.professionVersionsService.create(
+              professionVersion,
+              user,
+            );
+
+          this.professionVersionsService.archive(professionVersionToBeArchived);
```
bryangaledxw, line 195: I thin this should be `await`ed

src/organisations/organisation-versions.service.ts
```diff
@@ -176,6 +181,21 @@ export class OrganisationVersionsService {
         await this.repository.save(liveVersion);
       }
 
+      version.organisation.professions.forEach(async (profession) => {
```
bryangaledxw, line 184: I think this should be something like:

`await Promise.all(version.organisation.professions.map(async (profession) => {`

or maybe even a regular for loop to make sure the inner function is executed one-by-one for each profession without any interleaving

bryangaledxw (approved):
Looks good, just the `await`ing in `archive that needs sorting

PR #345, opened 2022-02-28
(2073) Handle user entered URLs not containing protocol

bryangaledxw:
We now silently add a default protocol if this is missing from the URL, and for any existing data in our database that is missing a URL protocol, we add the URL at presentation time.

### DIFF
--- a/src/helpers/format-link.helper.ts
+++ b/src/helpers/format-link.helper.ts
@@ -1,0 +1,19 @@
+import { isURL } from 'class-validator';
+import { escape } from './escape.helper';
+import { preprocessUrl, urlOptions } from './preprocess-url.helper';
+
+export function formatLink(url: string): string {
+  if (!url) {
+    return '';
+  }
+
+  const preprocessedUrl = preprocessUrl(url);
+
+  if (!isURL(preprocessedUrl, urlOptions)) {
+    return escape(url);
+  }
+
+  const escapedUrl = escape(preprocessedUrl);
+
+  return `<a href="${escapedUrl}" class="govuk-link">${escapedUrl}</a>`;
+}
--- a/src/helpers/validator.ts
+++ b/src/helpers/validator.ts
@@ -1,19 +1,22 @@
 import { validate, ValidationError } from 'class-validator';
 import { plainToClass } from 'class-transformer';
 
-export class Validator {
-  obj: object;
+export class Validator<T extends object> {
+  obj: T;
   dtoType: any;
   errors: ValidationError[];
 
-  public static async validate(dtoType: any, obj: object) {
-    const object: object = plainToClass(dtoType, obj);
+  public static async validate<T extends object>(
+    dtoType: { new (): T },
+    obj: object,
+  ) {
+    const object: T = plainToClass(dtoType, obj);
     const errors = await validate(object);
 
-    return new Validator(dtoType, obj, errors);
+    return new Validator<T>(dtoType, object, errors);
   }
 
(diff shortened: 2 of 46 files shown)


src/professions/admin/dto/registration.dto.ts
```diff
@@ -17,4 +17,6 @@ export class RegistrationDto {
   )
   @ValidateIf((e) => e.registrationUrl)
   registrationUrl: string;
+
+  change: string;
```
pezholio, line 24: Can we strongly type the contents of the string here? Maybe using a template literal?

src/organisations/admin/dto/organisation.dto.ts
```diff
@@ -18,20 +23,16 @@ export class OrganisationDto {
...
-  @IsUrl(
-    {},
-    {
-      message: 'organisations.admin.form.errors.url.invalid',
-    },
-  )
+  @IsUrl(urlOptions, {
+    message: 'organisations.admin.form.errors.url.invalid',
+  })
+  @Transform(({ value }) => preprocessUrl(value))
```
pezholio, line 29: 👍

src/helpers/validator.ts
```diff
@@ -1,19 +1,22 @@
...
   dtoType: any;
   errors: ValidationError[];
 
-  public static async validate(dtoType: any, obj: object) {
-    const object: object = plainToClass(dtoType, obj);
+  public static async validate<T extends object>(
+    dtoType: { new (): T },
+    obj: object,
+  ) {
+    const object: T = plainToClass(dtoType, obj);
```
pezholio, line 13: 👍

src/helpers/format-link.helper.ts
```diff
@@ -0,0 +1,19 @@
+import { isURL } from 'class-validator';
+import { escape } from './escape.helper';
+import { preprocessUrl, urlOptions } from './preprocess-url.helper';
+
+export function formatLink(url: string, classes = ''): string {
```
pezholio, line 5: Should we have `govuk-link` as the default here?

views/admin/professions/check-your-answers.njk
```diff
@@ -482,7 +482,7 @@
                     html: ("professions.form.label.legislation.link" | t) + "<span class=\"govuk-visually-hidden\"> " + loop.index + "</span>"
                   },
                   value: {
-                    html: ("<a class='govuk-link' href='" + (legislation.url | escape) + "'>" + (legislation.url | escape) + "</a>") if legislation else ""
+                    html: (legislation.url | link("govuk-link")) if legislation else ""
```
pezholio, line 485: Do we need this conditional here now?

pezholio (approved):
Lovely bit of work this - thanks! 👍 Just a couple of things, but nothing is a deal-breaker.

src/helpers/format-link.helper.ts
bryangaledxw: Let's!

views/admin/professions/check-your-answers.njk
bryangaledxw: We do not! I keep forgetting `.` in Nunjucks is basically `?.`

src/professions/admin/dto/registration.dto.ts
bryangaledxw: I've changed these to:

```
@Transform(({ value }) => parseBoolean(value))
change: boolean;
```

PR #360, opened 2022-03-02
(2034) Change HTML <title> attribute to match page title

pezholio:
This updates the templates to make sure the page titles are set in the HTML. I've added a `pageTitle` block to show the app name. If there is a `title` variable present, then we append this with a `|` character and spaces before.

The admin editing pages have the caption (which gives the name of edit/create action and the entity) and the title (telling the user where they are in the journey) again, seperated by pipes.

### DIFF
--- a/views/admin/professions/show.njk
+++ b/views/admin/professions/show.njk
@@ -3,6 +3,7 @@
 {% from "govuk/components/button/macro.njk" import govukButton %}
 
 {% set bodyClasses = "rpr-internal__page rpr-details__page" %}
+{% set title = profession.name %}
 
 {% block content %}
   {% include "../../_messages.njk" %}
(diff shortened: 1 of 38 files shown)


views/admin/professions/show.njk
```diff
@@ -3,6 +3,7 @@
 {% from "govuk/components/button/macro.njk" import govukButton %}
 
 {% set bodyClasses = "rpr-internal__page rpr-details__page" %}
+{% set title = (("professions.admin.status." + profession.status) | t) %}
```
Gweaton, line 6: This isn't quite right, you've set the title to the status in the `<aside>`, it should come from the `profession-details` template (the `<title> is `Regulated Professions Register | Live`)

Gweaton (approved):
Looks good, thanks for sorting this - just one thing but afterwards we're good to go!

views/admin/professions/show.njk
pezholio: Ooof, good catch!

PR #362, opened 2022-03-02
Fix domains

pezholio:
We had some teething issues with setting up domains on staging. These fixes should help us get prod set up quicker.


rjw1 (approved):
These seem like sensible changes.

PR #372, opened 2022-03-02
Add some documentation on adding a domain

pezholio:
We’ll probably need to do this again for the service domain, so this is a bit of an aide memoire, so we don’t forget how to do it!

### DIFF
--- a/doc/adding-a-domain.md
+++ b/doc/adding-a-domain.md
@@ -1,0 +1,172 @@
+# Adding a domain
+
+All of our infrastructure is managed with [Terraform](https://github.com/UKGovernmentBEIS/regulated-professions-register/tree/main/terraform),
+and in most cases, any infrastructure changes are applied automatically. There is one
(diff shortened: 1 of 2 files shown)


doc/adding-a-domain.md
```diff
@@ -0,0 +1,163 @@
...
+
+Assuming you have `tfenv` and `terraform` installed, make a copy of the
+`.env.ENVIRONMENT` file from the beis-rpr 1Password vault and put it in
+the `/terraform` directory. Then run:
+
+```bash
+source .env.ENVIRONMENT
+```
+
+To load the relevant environment variables that Terraform needs.
```
rjw1, line 94: we probably need to mention `terraform workspace select $ENV`

rjw1 (changes requested):
All looks good apart from my comment about `terraform workspace`

rjw1 (approved):

PR #375, opened 2022-03-03
(2181) Remove Mandatory Registration fields and radios

Gweaton:
# Changes in this PR

Removes the Mandatory Registration checkbox on the Registration page, as well as on the Profession & ProfessionVersion entities.

Apologies, this is all one commit, but I wanted to make sure all tests were passing - in essence, it's doing the following:
- Removing the fields (and MandatoryRegistration Enum) from the entities
- Removing the radio buttons from the Registration page
- Removing the MandatoryRegistrationCheckboxPresenter
- Not displaying the MandatoryRegistration value on the Check your answers page.

### Before

image

image

### After

image

image


pezholio (approved):
LGTM 👍

PR #376, opened 2022-03-03
Redirect to canonical hostname

pezholio:
Now we have a domain set up, this adds some middleware to ensure that all requests are redirected to the correct domain.

### DIFF
--- a/src/middleware/redirect-to-canonical-hostname.spec.ts
+++ b/src/middleware/redirect-to-canonical-hostname.spec.ts
@@ -1,0 +1,64 @@
+import { createMock, DeepMocked } from '@golevelup/ts-jest';
+import { Request, Response, NextFunction } from 'express';
+import { redirectToCanonicalHostname } from './redirect-to-canonical-hostname';
+
+describe('redirectToCanonicalHostname', () => {
+  let request: DeepMocked<Request>;
+  let response: DeepMocked<Response>;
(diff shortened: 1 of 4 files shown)


src/middleware/redirect-to-canonical-hostname.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+  let next: jest.Mock<NextFunction>;
+  let host: string;
+
+  const OLD_ENV = process.env;
+
+  beforeEach(async () => {
+    jest.resetModules();
+    process.env = { ...OLD_ENV, CANONICAL_HOSTNAME: 'www.example.org' };
+
+    request = createMock<Request>({
```
Gweaton, line 17: As a reader of this test, I'd find it a bit more clear if the `request` and `reponse` objects in this `beforeEach` were inside the `it` block with the test itself, as I have to scroll up to find where they're set. How would you feel about moving it into the `it` body (along with the setup in the  `beforeEach` nested in the `describe` below?

Gweaton (approved):
Looks good, just a comment about test structure

src/middleware/redirect-to-canonical-hostname.spec.ts
pezholio: No problemo 👍

PR #377, opened 2022-03-03
(2183) Remove no-longer-needed fields from Qualifications

Gweaton:
# Changes in this PR

Removes no-longer-needed fields from Qualifications:

- Qualification level
- Most common route to obtain
- Education duration
- Existence of professional experience required

We're now not displaying these Qualification fields anywhere e.g. on the
"show" pages or the Check your answers page when adding/editing a
profession.

I've had to update the seeds here, as we can no longer rely on a
qualification level to find the relevant qualification to set on a
profession, so I've gone for using the `routeToObtain` value here.

#### Qualification editing page

image

#### Check your answers

image

#### Public facing page

image

#### Qualification editing page

image

#### Check your answers

image

#### Public facing page

image


src/professions/professions.seeder.ts
```diff
@@ -133,14 +133,7 @@ export class ProfessionsSeeder implements Seeder {
...
           qualification = await this.qualificationsRepository.save(
-            new Qualification(
-              qualification.level,
-              qualification.routesToObtain,
-              qualification.mostCommonRouteToObtain,
-              qualification.educationDuration,
-              qualification.mandatoryProfessionalExperience,
-              qualification.url,
-            ),
+            new Qualification(qualification.routesToObtain, qualification.url),
```
pezholio, line 136: Do we even need the Qualifications relation now? Might simplify things to just have two fields in the Profession.

pezholio (approved):
Love to see some deleted code! This is good to go, but just one thing about qualifications. I'm tempted to say leave it until another time, but it might be good to do it now while we're touching this bit of the codebase. What do you think?

src/professions/professions.seeder.ts
Gweaton: There are some more fields like `ukRecognition` `otherCountriesRecognition` which are optional so we're not seeding them at the moment, so I think it still makes sense to have a separate Qualification (as much as I'd love to have fewer database entities).

pezholio: Ah course!

PR #385, opened 2022-03-04
(2199) Second regulator added doesn't appear on page (part 2)

bryangaledxw:
# Changes in this PR

We now display multiple Organisations for a Profession when displayed in internal and external search listings, completing 2199. We also now inspect the additional Organisation when searching Professions by Organisation

### Before
localhost_3000_professions_search (2)
localhost_3000_admin_professions (2)

### After
localhost_3000_professions_search (1)
localhost_3000_admin_professions (1)

### DIFF
--- a/src/professions/admin/profession-versions.controller.spec.ts
+++ b/src/professions/admin/profession-versions.controller.spec.ts
@@ -19,6 +19,7 @@ import { ProfessionPresenter } from '../presenters/profession.presenter';
 import { getActingUser } from '../../users/helpers/get-acting-user.helper';
 import { createDefaultMockRequest } from '../../testutils/factories/create-default-mock-request';
 import organisationFactory from '../../testutils/factories/organisation';
+import * as getOrganisationsFromProfessionModule from '../helpers/get-organisations-from-profession.helper';
 
 jest.mock('../../organisations/organisation.entity');
 jest.mock('../presenters/profession.presenter');
@@ -125,6 +126,11 @@ describe('ProfessionVersionsController', () => {
             (organisation) => organisation,
           );
 
+          const getOrganisationsFromProfessionSpy = jest.spyOn(
+            getOrganisationsFromProfessionModule,
+            'getOrganisationsFromProfession',
+          );
+
           const result = await controller.show('profession-id', 'version-id');
 
           expect(result).toEqual({
@@ -142,6 +148,9 @@ describe('ProfessionVersionsController', () => {
           expect(
             professionVersionsService.findByIdWithProfession,
           ).toHaveBeenCalledWith('profession-id', 'version-id');
+          expect(getOrganisationsFromProfessionSpy).toHaveBeenCalledWith(
+            professionWithVersion,
+          );
         });
       });
 
@@ -172,6 +181,11 @@ describe('ProfessionVersionsController', () => {
             (organisation) => organisation,
           );
 
+          const getOrganisationsFromProfessionSpy = jest.spyOn(
+            getOrganisationsFromProfessionModule,
+            'getOrganisationsFromProfession',
+          );
+
           const result = await controller.show('profession-id', 'version-id');
 
           expect(result).toEqual({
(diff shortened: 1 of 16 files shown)


src/professions/helpers/get-organisations-from-profession.helper.spec.ts
```diff
@@ -0,0 +1,51 @@
+import organisationFactory from '../../testutils/factories/organisation';
+import professionFactory from '../../testutils/factories/profession';
+import { getOrganisationsFromProfession } from './get-organisations-from-profession.helper';
+
+describe('getOrganisationsFromProfession', () => {
+  describe('when the Prganisation has ony a main Organisation', () => {
```
Gweaton, line 6: typos in `Organisation` and `only` here (sorry!)

src/professions/helpers/get-organisations-from-profession.helper.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+        organisation,
+      });
+
+      expect(getOrganisationsFromProfession(profession)).toEqual([
+        organisation,
+      ]);
+    });
+  });
+
+  describe('when the Prganisation has a main Organisation and an additional Organisation', () => {
```
Gweaton, line 20: another typo on the first `organisation` here

src/professions/helpers/get-organisations-from-profession.helper.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+      });
+
+      expect(getOrganisationsFromProfession(profession)).toEqual([
+        organisation,
+        additionalOrganisation,
+      ]);
+    });
+  });
+
+  describe('when the Prganisation has ony an additional Organisation', () => {
```
Gweaton, line 37: one more here

src/professions/helpers/get-organisations-from-profession.helper.ts
```diff
@@ -0,0 +1,21 @@
+import { Organisation } from '../../organisations/organisation.entity';
+import { Profession } from '../profession.entity';
+
+export function getOrganisationsFromProfession(
```
Gweaton, line 4: Did you consider making this a static method on the `Profession` entity like we have `Profession.withVersion`? Do you think it could be a sensible place for it to live too?

src/professions/admin/profession-versions.controller.spec.ts
```diff
@@ -287,5 +304,6 @@ describe('ProfessionVersionsController', () => {
 
   afterEach(() => {
     jest.resetAllMocks();
+    jest.restoreAllMocks();
```
Gweaton, line 307: I didn't know about this - is this needed in addition to `resetAllMocks` or can it replace it?

views/professions/search/index.njk
```diff
@@ -39,6 +39,12 @@
             {% endfor %}
           {% endset %}
 
+          {% set organisationsHtml %}
+            {% for organisation in profession.organisations %}
+              <p class="govuk-body-m">{{ organisation }}</p>
```
Gweaton, line 44: Do we need to escape `organisation` here as we're outputting HTML?

Gweaton (approved):
This looks good to me, have just left a couple of comments (including some about typos) and one about where to put the new helper function, let me know what you think.

src/professions/admin/profession-versions.controller.spec.ts
bryangaledxw: Afraid we need both - going by the docs, this only works for mocks created via `jest.spyOn`

views/professions/search/index.njk
bryangaledxw: We do not! Nunjucks escapes by default - where we need to worry about escaping is where we've explicitly said "expect this to be HTML", by say, using the "html" attribute on GDS components

src/professions/helpers/get-organisations-from-profession.helper.ts
bryangaledxw: I did but my leaning was that if it didn't need to be in there, keep it as its own isolated thing. It's not like `withVersion` which needs to change as `Profession` itself changes. But can move if you'd rather

Gweaton: I was thinking having it as part of the `Profession` would sort of be a way of scoping and making it clear what its responsibilities are, but I the function name you've chosen does that well enough anyway. Happy to leave it where it is.

PR #395, opened 2022-03-07
(2216) Audit optional fields and a couple of related fixes

Gweaton:
# Changes in this PR

- Fixes error not being displayed when not selecting a nation.
- Fixes incorrect `<title>` on page.
- Ensures all optional fields when adding a profession are actually optional and marks them as so.
- Only validate legislation link if previous step in form is filled in.
- Updates a couple of tests to make sure we can display professions with only mandatory fields filled in.

### After

image

image

image

image

image

image


bryangaledxw (approved):
Looks good. Really helpful for the stuff I'm working on to have those new tests already in place

PR #397, opened 2022-03-08
(2226) Escape names passed in to flash message

Gweaton:
# Changes in this PR

Ensures all names passed into flash message are escaped to prevent possibility of XSS attacks via the confirmation message banner.

I initially tried to escape this in the banner itself, but because we use HTML in some of our translated strings, this was outputting `"<strong>"` on the page, rather than the strong tag.

### DIFF
--- a/src/organisations/admin/organisation-archive.controller.spec.ts
+++ b/src/organisations/admin/organisation-archive.controller.spec.ts
@@ -3,6 +3,7 @@ import { TestingModule, Test } from '@nestjs/testing';
 import { Response } from 'express';
 import { I18nService } from 'nestjs-i18n';
 import { flashMessage } from '../../common/flash-message';
+import { escape } from '../../helpers/escape.helper';
 import { createMockI18nService } from '../../testutils/create-mock-i18n-service';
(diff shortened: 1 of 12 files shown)


src/organisations/admin/organisation-archive.controller.spec.ts
```diff
@@ -81,6 +83,7 @@ describe('OrganisationArchiveController', () => {
 
       const res = createMock<Response>({});
 
+      (escape as jest.Mock).mockImplementation();
```
bryangaledxw, line 86: Does this do anything if we're not actually providing a mock implementation?

bryangaledxw: Does `jest.spyOn` work just as well, since if we just want to check something is being call, rather than provide a mock implementation?

bryangaledxw (approved):

src/organisations/admin/organisation-archive.controller.spec.ts
Gweaton: Good point, these weren't actually necessary because we're mocking the imports at the top of the file. I've removed them.

PR #403, opened 2022-03-08
Make regulation type an enum

pezholio:
As the `regulationType` can only be one of a limited set of values, it makes sense (and will help when we eventually make it searchable) for it to be an enum


Gweaton (approved):
Looks good to me

PR #405, opened 2022-03-09
(2140) UI iterations to internal professions details page

bryangaledxw:
# Changes in this PR

* Empty fields are shown in internal Organisation and Profession pages and hidden in public pages
* Page status are shown using a tag
* Internal pages show a link to the public page
* Buttons on internal pages are re-ordered
* We display both the name and the email address of the last editor of a page

### Before
localhost_3000_professions_no-optional-fields (1)
localhost_3000_admin_professions_74faa7be-e598-4c20-b00e-04e1303e0a91_versions_c5d275dc-c29f-4ebc-84c5-3f32e55bc4e2 (1)

### After
localhost_3000_professions_no-optional-fields
localhost_3000_admin_professions_74faa7be-e598-4c20-b00e-04e1303e0a91_versions_c5d275dc-c29f-4ebc-84c5-3f32e55bc4e2


views/professions/_profession-details.njk
```diff
@@ -58,31 +58,33 @@
           value: {
             html: organisation.address | multiline
           }
-        },
+        } if (organisation.address or showEmptyProfessionDetails) else undefined,
```
Gweaton, line 61: I find this a little hard to read - is it possible to put the whole address etc in an `if` statement (rather than having the `if` at the end, or is that not possible with nunjucks?

src/organisations/presenters/organisation.presenter.spec.ts
```diff
@@ -216,6 +216,9 @@ describe('OrganisationPresenter', () => {
         it('should filter out empty fields', async () => {
           const i18nService = createMockI18nService();
           (escape as jest.Mock).mockImplementation(escapeOf);
+          (formatMultilineString as jest.Mock).mockImplementation(multilineOf);
+          (formatEmail as jest.Mock).mockImplementation(emailOf);
```
Gweaton, line 228: Are these needed here, or were they just missing in the previous implementation?

src/helpers/pad.helper.ts
```diff
@@ -0,0 +1,13 @@
+export function pad<T>(array: T[], minimumLength: number): T[] {
```
Gweaton, line 1: Do you think you could add a bit more detail to the commit message about what this is used for?

views/professions/_profession-details.njk
Gweaton: Ideally, we'd construct this in a presenter and output the values based on what values are there, then we could pass it into the page and avoid using some slightly obscure `undefined` behaviour. In the interests of time, I think this is ok for now, but it would be good to add a card to tidy this up later.

Gweaton (approved):
Thanks for all this, looks like a lot of work.

Have reviewed as best I can, but as this was a rather long PR, I found it a bit tricky to spot everything. It would be easier in terms of reviewing to have it split up a little bit, but I appreciate that would come with some overhead on your part.

src/organisations/presenters/organisation.presenter.spec.ts
bryangaledxw: Yep, otherwise some fields return `undefined` and get wrongly stripped out (now that we're checking for falsey values rather than specifically `''`)

views/professions/_profession-details.njk
bryangaledxw: Added 2242. Yeah, without adding a new presenter, this is the way to do it. Nunjucks won't let you add an `if` block mid object declaration.

Though I did just spot a couple of redundant `undefined`s (I think i'd chucked some in to look at the HTML it generated) that i've now revoed

bryangaledxw:
Thanks for the feedback. Apologies this turned into an epic! I underestimated how the individually quite small changes would add up!

PR #406, opened 2022-03-09
(2169) Update sectors

Gweaton:
# Changes in this PR

- Add Environment and Food sector
- Rename Contruction & Engineering to Construction and Engineering
- Rename Health to Health and Care
- Move "Others" to bottom of the list

### After

image

image

### DIFF
--- a/seeds/test/industries.json
+++ b/seeds/test/industries.json
@@ -9,6 +9,9 @@
     "name": "industries.education"
   },
   {
+    "name": "industries.environmentAndFood"
+  },
+  {
     "name": "industries.finance"
   },
   {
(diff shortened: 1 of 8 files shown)


seeds/test/industries.json
```diff
@@ -14,13 +17,13 @@
...
-    "name": "industries.other"
-  },
   {
     "name": "industries.security"
   },
   {
     "name": "industries.transport"
+  },
+  {
+    "name": "industries.other"
```
pezholio, line 27: I'm a bit conscious we're relying on the file order to make this the last one. I'd feel a bit more comfortable if we were specific at the code level which one we want as the "last" one.

pezholio (approved):
LGTM 👍

PR #407, opened 2022-03-09
Add missing industries

pezholio:
The professions data we’re preparing to import has some industries that we’re missing. This adds them, along with translations.


Gweaton (approved):
You might want to pop this on my other branch [in the other PR](https://github.com/UKGovernmentBEIS/regulated-professions-register/pull/406) open!

PR #413, opened 2022-03-09
(2239) Hide name field when submitting org with errors

Gweaton:
# Changes in this PR

Before this change, a slug value wouldn't be sent on a form submittion
error due to the way we use a `ValidationExceptionFilter` here (rather
than passing the values through like we do on the Profession side of
things).

On an error, the page is reloaded, but as the `slug` value is null,
`slug|length == 0` always evaluates to `true` and the `name` field is
shown.

This approach is a bit of a quick, hacky fix - as I think ideally we'd
either look up the Organisation from the database and handle this like
we do for Professions (which is a big change and feels a bit risky to do
at this stage), or modify the `ValidationExceptionFilter` to look up the
entity for us (which I haven't been able to do yet).

I've had to add a `submittedName` field here to also correctly render
the `name` of the organisation, otherwise that comes through as blank.

### Before

image

### After

image

### DIFF
--- a/views/admin/organisations/edit.njk
+++ b/views/admin/organisations/edit.njk
@@ -3,6 +3,8 @@
 {% from "govuk/components/textarea/macro.njk" import govukTextarea %}
 {% from "govuk/components/button/macro.njk" import govukButton %}
 
+{% set name = name or submittedName %}
+
 {% if slug|length == 0 %}
   {% set title = ("organisations.admin.create.heading" | t) %}
 {% else %}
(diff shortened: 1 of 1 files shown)


views/admin/organisations/edit.njk
```diff
@@ -6,7 +6,7 @@
 {% if slug|length == 0 %}
   {% set title = ("organisations.admin.create.heading" | t) %}
 {% else %}
-  {% set title = ("organisations.admin.edit.heading" | t({ organisationName: name })) %}
+  {% set title = ("organisations.admin.edit.heading" | t({ organisationName: name or submittedName })) %}
```
bryangaledxw, line 9: Would it be better to leave this line as is, and do `{% set name = name or subittedName %}` at the very top of the page? Then we get the correct value in the first `govInput`too.

Or if not, does the first `govInput` need `value: name or submittedName`?

bryangaledxw (approved):

views/admin/organisations/edit.njk
Gweaton: Very good point - I think that's a lot nicer. Have fixed that up (the first suggestion)

PR #414, opened 2022-03-10
(2237) Display organisation name on admin dashboard

Gweaton:
# Changes in this PR

- Displays user's organisation (or BEIS for service owners) on dashboard

### Before

image

#### BEIS user

image

#### Organisation user

image

### DIFF
--- a/src/app.controller.ts
+++ b/src/app.controller.ts
@@ -1,5 +1,8 @@
-import { Controller, Get, Render, UseGuards } from '@nestjs/common';
+import { Controller, Get, Render, Req, UseGuards } from '@nestjs/common';
 import { AuthenticationGuard } from './common/authentication.guard';
+import { RequestWithAppSession } from './common/interfaces/request-with-app-session.interface';
+import { getActingUser } from './users/helpers/get-acting-user.helper';
+import { getUserOrganisation } from './users/helpers/get-user-organisation';
 @Controller()
 export class AppController {
   @Get()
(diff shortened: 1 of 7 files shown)


src/app.controller.ts
```diff
@@ -19,6 +19,7 @@ export class AppController {
 
     return {
       organisation: getUserOrganisation(actingUser),
+      isBEISUser: actingUser.serviceOwner,
```
bryangaledxw, line 22: I think we already get this? The user is available to all templates via `globalLocals`, so you can do {% if user.serviceOwner %} directly

views/admin/dashboard.njk
```diff
@@ -29,6 +29,7 @@
     </nav>
   </div>
   <div class="govuk-grid-column-two-thirds">
+    <span class="govuk-caption-l">{{ organisation | t }}</span>
```
bryangaledxw, line 32: I'm not a fan of passing something into the `t` filter that might not actually need translating, I'd be happier if the translation happened within `getUserOrganisation` in the case where the use is a service owner. If we ever want automated checking to see if unrecognised string IDs are being passed into the translator, this'll defeat that

I think this is fine for now, but would like a ticket to address

pezholio (approved):
LGTM - just one thought - do we want to make `isBeisUser` a global, so we can use it everywhere?

src/app.controller.ts
bryangaledxw: Also, do we want to mix terminology in code? I feel like I'd rather keep to `serviceOwner`, and references to "BEIS" live in the translation file

bryangaledxw (approved):

src/app.controller.ts
Gweaton: Ah, sorry I didn't realise we had this in the globals already - have removed this commit.

views/admin/dashboard.njk
Gweaton: Yeah, I had the same concerns with this, but this was more of a quick fix. I'll add a ticket to address this later.

PR #417, opened 2022-03-10
Send the request to Rollbar

pezholio:
Initially, we weren't sending any data about the request when an error occurred, so it's difficult to track down exactly what's caused an error. I've also ensured any IP addresses aren't sent with errors, to keep us in line with GDPR.

### Before

image

### After

image


Gweaton (approved):
Nice one, thanks so much for sorting this

PR #424, opened 2022-03-11
Ensure profession pages render if occupation locations and industry data is missing

Gweaton:
# Changes in this PR

- Adds safety net for industries and occupation locations not being set on a Profession, as we don't currently have pre-publish validation. This will at least stop a broken experience for members of the public, and is a temporary stopgap until we implement the pre-publish validation.
- Doesn't render regulation heading when no regulation data is set.

### Before

[ERROR PAGE]

### After

image


pezholio (approved):
LGTM 👍

PR #425, opened 2022-03-11
Add docs for clearing the database before re-import

Gweaton:
# Changes in this PR

Adds instructions for wiping the database before a data import.

We'll be doing this to import data to Production, and it's a good idea
to have the instructions captured here in case we ever need to do it
again.


pezholio (approved):
LGTM - I think Prettier doesn't like those code fences though

PR #434, opened 2022-03-14
(2204) Index Professions in Opensearch when publishing

pezholio:
This is the first part of the improving search puzzle. Now we have ~~Elasticsearch~~ Opensearch in place, this puts all the relevant pieces into place and indexes professions as soon as they are published, as well as deindexing older versions. The next PR will work on using the indexes in the public-facing search.


src/professions/profession-versions.service.ts
```diff
@@ -100,6 +102,8 @@ export class ProfessionVersionsService {
       await queryRunner.release();
     }
 
+    await this.searchService.index(version);
```
Gweaton, line 106: Do we want to do this as part of the transaction, or is it safe to do this outside?

docker-compose.ci.yml
```diff
@@ -61,6 +62,16 @@ services:
     volumes:
       - redis-data:/var/lib/redis
 
+  opensearch:
+    image: 'opensearchproject/opensearch'
+    environment:
+      discovery.type: single-node
+      plugins.security.disabled: 1
```
Gweaton, line 69: what does this mean?

src/seeder.ts
```diff
@@ -43,7 +46,9 @@ seeder({
       Profession,
       ProfessionVersion,
     ]),
+    SearchModule.register(),
   ],
+  providers: [ProfessionsSearchService],
```
Gweaton, line 51: Do we need to re-run the seeds on Production, then?

Gweaton (approved):
I think this makes sense to me, thanks for putting all of this in place!

Just have a couple of questions, really:

src/professions/profession-versions.service.ts
```diff
@@ -90,6 +90,7 @@ export class ProfessionVersionsService {
       if (liveVersion) {
         liveVersion.status = ProfessionVersionStatus.Archived;
         await this.repository.save(liveVersion);
+        await this.searchService.delete(liveVersion);
```
bryangaledxw, line 93: We should also do this in `ProfessionVersionsService.archive`, right? Also the commit message on this one feels misleading - "Delete previously indexed version when archiving" instead?

src/professions/professions-search.service.ts
```diff
@@ -4,12 +4,14 @@ import { ProfessionVersion } from './profession-version.entity';
 
 @Injectable()
 export class ProfessionsSearchService {
+  indexName: string = `professions_${process.env['NODE_ENV']}`;
```
bryangaledxw, line 7: Suggest make `private readonly`

src/professions/professions.seeder.ts
```diff
@@ -101,6 +104,14 @@ export class ProfessionsSeeder implements Seeder {
         ).filter(Boolean);
 
         await this.professionVersionsRepository.save(seededVersions);
+
+        const liveVersions = seededVersions.filter(
+          (version) => version.status === ProfessionVersionStatus.Live,
+        );
+
+        for (const version of liveVersions) {
+          await this.searchService.index(version);
```
bryangaledxw, line 113: Do we also need to delete the professions from opensearch within `drop`?

bryangaledxw (approved):
Looks good, with the caveat that I'm needing to take the docker/CI parts on trust!

src/professions/profession-versions.service.ts
pezholio: I think it's safe, because we won't be able to reverse the index if the transaction fails, so we want to index after the fact

docker-compose.ci.yml
pezholio: So we're not running over https

src/seeder.ts
pezholio: No, I think we'll need to add a task to index all the live content

src/professions/profession-versions.service.ts
pezholio: I'm gonna address this in the next PR, as it's a bit more complex 👍

PR #449, opened 2022-03-16
(2115) Improve user editing constraints

Gweaton:
# Changes in this PR

Before this change, users could theoretically edit other users if they had a url containing a UUID of that user. This prevents that with a new `checkCanViewOrganisationData` function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Search profession titles using Opensearch
 - Allow central users to filter Profession and Organisation listings by regulation type
 
+### Changed
+
+- Improve constraints around editing users
+
 ## [release-009] - 2022-03-11
 
 ### Changed
--- a/src/users/helpers/check-can-view-organisation.spec.ts
+++ b/src/users/helpers/check-can-view-organisation.spec.ts
@@ -1,0 +1,99 @@
+import { UnauthorizedException } from '@nestjs/common';
+import { createDefaultMockRequest } from '../../testutils/factories/create-default-mock-request';
+import organisationFactory from '../../testutils/factories/organisation';
+import userFactory from '../../testutils/factories/user';
+import { checkCanViewOrganisation } from './check-can-view-organisation';
+import { getActingUser } from './get-acting-user.helper';
+
+jest.mock('./get-acting-user.helper');
+
+describe('checkCanViewOrganisation', () => {
+  describe('when called with a request from a non-service owner user', () => {
+    describe('when trying to view data from the same organisation', () => {
+      it('returns without throwing an Error', () => {
+        const organisation = organisationFactory.build();
+        const user = userFactory.build({
+          serviceOwner: false,
+          organisation: organisation,
+        });
+
+        (getActingUser as jest.Mock).mockReturnValue(user);
+
+        const request = createDefaultMockRequest({ user: user });
+
+        expect(() => {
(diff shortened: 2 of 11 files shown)


src/users/helpers/check-can-view-organisation-data.ts
```diff
@@ -0,0 +1,17 @@
+import { UnauthorizedException } from '@nestjs/common/exceptions/unauthorized.exception';
+import { Organisation } from '../../organisations/organisation.entity';
+import { User } from '../user.entity';
+
+export function checkCanViewOrganisationData(
+  currentUser: User,
+  organisation?: Organisation | null,
+): boolean {
```
bryangaledxw, line 8: Think this should be `void`, and the function either throws or doesn't throw, rather than returning anything

src/users/helpers/check-can-view-organisation-data.ts
```diff
@@ -0,0 +1,17 @@
+import { UnauthorizedException } from '@nestjs/common/exceptions/unauthorized.exception';
+import { Organisation } from '../../organisations/organisation.entity';
+import { User } from '../user.entity';
+
+export function checkCanViewOrganisationData(
+  currentUser: User,
```
bryangaledxw, line 6: I'd be tempted to have this take a `Request`, and this function does the call to `getActingUser`, rather than every user of this function have to do that.

Or if we're keeping this, I'd rename `currentUser` to `actingUser` for consistency

src/users/helpers/check-can-view-organisation-data.ts
```diff
@@ -0,0 +1,17 @@
...
+  organisation?: Organisation | null,
+): boolean {
+  if (
+    currentUser.serviceOwner ||
+    currentUser.organisation?.id === organisation?.id
```
bryangaledxw, line 11: I feel like err-ing on the side of saftey, `throw` if the organisation null/undefined. Here, if both are `null` that's probably a mistake, but this'll return true

src/users/helpers/check-can-view-organisation-data.ts
```diff
@@ -0,0 +1,17 @@
+import { UnauthorizedException } from '@nestjs/common/exceptions/unauthorized.exception';
+import { Organisation } from '../../organisations/organisation.entity';
+import { User } from '../user.entity';
+
+export function checkCanViewOrganisationData(
+  currentUser: User,
+  organisation?: Organisation | null,
```
bryangaledxw, line 7: I'd suggest do a `checkCanViewUser` (and later, `checkCanViewProfession`) that calls down to this function, rather than use this directly in this PR

bryangaledxw (approved):
I've gone into very fiddly detail on the new function, but think this is the right approach.

I was trying to work out how you might do it in a decorator - I think it would have to involve inspecting the request URL to figure out what user/org/profession we're viewing, and do additional duplicated DB requests. This feels like the more explicit was to do it

src/users/helpers/check-can-view-organisation-data.ts
Gweaton: So, a central admin doesn't actually have an organisation - I don't know if that's by design or not - but I suppose `currentUser.organisation` should never be `null` here, as it'll be caught by the first part of `||`.

bryangaledxw: Ah, yep, read my comment as applying to everything *after* the `currentUser.serviceOwner` check

bryangaledxw (approved):

PR #450, opened 2022-03-16
(2204) Search Profession titles using opensearch

pezholio:
This pulls together the previous Opensearch-related PRs to add an endpoint that runs a search query against indexed data in Opensearch. We then have a database-level search method that gets the IDs of the versions that match a query and adds any other filters to it.

At the moment, this just replicates the functionality we already have. The next job is to index and search against other fields.

### DIFF
--- a/src/professions/professions-search.service.ts
+++ b/src/professions/professions-search.service.ts
@@ -1,7 +1,10 @@
 import { OpensearchClient } from 'nestjs-opensearch';
 import { Injectable } from '@nestjs/common';
+import {
+  SearchResponse,
+  SearchHit,
+} from '@opensearch-project/opensearch/api/types';
 import { ProfessionVersion } from './profession-version.entity';
-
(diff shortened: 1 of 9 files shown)


src/professions/professions-search.service.ts
```diff
@@ -39,4 +42,22 @@ export class ProfessionsSearchService {
       },
     });
   }
+
+  public async search(query: string): Promise<String[]> {
```
bryangaledxw, line 46: Can this be `Promise<string>` for consistency?

pezholio: Do you mean `Promise<string[]>`?

bryangaledxw: Sorry yes! Just being picking about `String` vs. `string`

pezholio: Cool - thought that was the case 😃

PR #456, opened 2022-03-16
(2274) Improve Organisation editing constraints

Gweaton:
# Changes in this PR

Ensures users can't view other Organisation admin pages unless they're central admins. They could only do this if they had the UUID of the organisation and the version, so it was very unlikely that they'd have been able to do this anyway.

### DIFF
--- a/src/organisations/admin/organisation-versions.controller.ts
+++ b/src/organisations/admin/organisation-versions.controller.ts
@@ -28,6 +28,7 @@ import { Permissions } from '../../common/permissions.decorator';
 import { UserPermission } from '../../users/user-permission';
 
 import { getActingUser } from '../../users/helpers/get-acting-user.helper';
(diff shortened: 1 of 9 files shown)


src/organisations/admin/organisation-versions.controller.ts
```diff
@@ -83,6 +85,8 @@ export class OrganisationVersionsController {
       true,
     );
 
+    checkCanViewOrganisation(req, organisation);
```
bryangaledxw, line 88: Would also put this check on the `@Post` endpoint

bryangaledxw (approved):

src/organisations/admin/organisation-versions.controller.ts
Gweaton: Ah yeah, well spotted. I got confused between creating a version (which only authorized users can do) and creating an Organisation (which anyone can do)

PR #457, opened 2022-03-17
(2204) Add a script to reseed search data

pezholio:
This adds a new npm script to delete all professions data from an index and reindex it all. I've also added some documentation, as well as an ADR, as we hadn't yet added one.


Gweaton (approved):
Looks good - we might want to make `ts-node` a prod dependency as we've been discussing on Slack, but we can also workaround that if you think that should come in its own PR.

PR #458, opened 2022-03-17
(2251) Add contact details to error pages

pezholio:
This adds contact details to the 404 and 500 pages to match the [pattern in the design system](https://design-system.service.gov.uk/patterns/problem-with-the-service-pages/). I've also added the contact details to a partial so we only have to update them in once place.

### Before

image

image

### After

image

image


views/shared/_contact-details.njk
```diff
@@ -0,0 +1,22 @@
...
+<p class="govuk-body">
+  Telephone<br>
+  <strong class="govuk-!-font-weight-bold">
+    <a href="tel:+442072155000" class="govuk-link">020 7215 5000</a>
+  </strong>
+</p>
+
+<p class="govuk-body">
+  Opening times:<br>
+  <strong class="govuk-!-font-weight-bold">Monday to Friday, 9am to 4:30pm</strong>
```
Gweaton, line 17: We've not done this for cookie page or privacy policy, but should these all be translations?

Gweaton (approved):
Looks good, thanks for sorting this, just one thing re: translated text

views/shared/_contact-details.njk
pezholio: I don't know if it's worth going through all the hassle at this point. It was probably overkill for the error messages.

PR #460, opened 2022-03-17
(2275) Improve constraints around setting Professions

Gweaton:
# Changes in this PR

Ensures users can't view other Profession admin pages or edit Professions from another Organisation unless they're central admins. They could only do this if they had the UUID of the Profession and the version, so it was very unlikely that they'd have been able to do this anyway.

### DIFF
--- a/src/users/helpers/check-can-view-profession.ts
+++ b/src/users/helpers/check-can-view-profession.ts
@@ -1,0 +1,10 @@
+import { RequestWithAppSession } from '../../common/interfaces/request-with-app-session.interface';
+import { Profession } from '../../professions/profession.entity';
+import { checkCanViewOrganisation } from './check-can-view-organisation';
+
+export function checkCanViewProfession(
+  request: RequestWithAppSession,
+  profession: Profession,
+) {
+  checkCanViewOrganisation(request, profession.organisation);
+}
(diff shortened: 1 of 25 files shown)


src/users/helpers/check-can-view-profession.ts
```diff
@@ -0,0 +1,10 @@
+import { RequestWithAppSession } from '../../common/interfaces/request-with-app-session.interface';
+import { Profession } from '../../professions/profession.entity';
+import { checkCanViewOrganisation } from './check-can-view-organisation';
+
+export function checkCanViewProfession(
+  request: RequestWithAppSession,
+  profession: Profession,
+) {
+  checkCanViewOrganisation(request, profession.organisation);
+}
```
pezholio, line 10: I'm not quite on board with the wording here. Would `checkUserPermissions` be better?

pezholio (approved):
LGTM - just a comment about naming, which I guess covers `checkCanViewOrganisation` as well

src/users/helpers/check-can-view-profession.ts
Gweaton: We've done this the same way with permissions for viewing Users, Organisations and Professions, so this is consistent for now, but we can probably rework this once this is all in. Ideally, we'd have some kind of guard that does this, but we didn't have a lot of time to spend on this - I'll add a card to tidy this up for the dev improvements if you're happy with this going in as-is for now?

PR #463, opened 2022-03-17
(2151) Replace per page validation with pre-publish validation

bryangaledxw:
# Changes in this PR
* Change Profession creation flow so that the user is taken to Check Your Answers after completing the top-level information
* Enforce that a Profession cannot be published until required sections are completed

### Before
localhost_3000_admin_professions_61997465-2d80-4d30-b059-960d05978337_versions_70a7ce54-3fa3-486e-9325-15a5c38b43b9 (2)
localhost_3000_admin_professions_61997465-2d80-4d30-b059-960d05978337_versions_cc3d6159-5133-4bec-a9c2-af3445586a05_check-your-answers

### After
localhost_3000_admin_professions_61997465-2d80-4d30-b059-960d05978337_versions_70a7ce54-3fa3-486e-9325-15a5c38b43b9 (1)
localhost_3000_admin_professions_61997465-2d80-4d30-b059-960d05978337_versions_d2282cc4-3f0a-47cb-8604-98b9facd7be3_check-your-answers

### DIFF
--- a/src/professions/admin/legislation.controller.ts
+++ b/src/professions/admin/legislation.controller.ts
@@ -4,12 +4,11 @@ import {
   Get,
   Param,
   Post,
-  Query,
   Req,
   Res,
   UseGuards,
 } from '@nestjs/common';
-import { Response, Request } from 'express';
+import { Response } from 'express';
 import { AuthenticationGuard } from '../../common/authentication.guard';
 import { Permissions } from '../../common/permissions.decorator';
 import { Validator } from '../../helpers/validator';
@@ -38,16 +37,13 @@ export class LegislationController {
 
   @Get('/:professionId/versions/:versionId/legislation/edit')
   @Permissions(UserPermission.CreateProfession, UserPermission.EditProfession)
-  @BackLink((request: Request) =>
-    request.query.change === 'true'
-      ? '/admin/professions/:professionId/versions/:versionId/check-your-answers'
-      : '/admin/professions/:professionId/versions/:versionId/qualifications/edit',
+  @BackLink(
+    '/admin/professions/:professionId/versions/:versionId/check-your-answers',
   )
   async edit(
     @Res() res: Response,
     @Param('professionId') professionId: string,
     @Param('versionId') versionId: string,
-    @Query('change') change: string,
     @Req() request: RequestWithAppSession,
   ): Promise<void> {
     const profession = await this.professionsService.findWithVersions(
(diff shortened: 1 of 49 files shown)


views/shared/_publication-blockers.njk
```diff
@@ -0,0 +1,16 @@
+{% if publicationBlockers.length %}
+<div>
+  <h2 class="govuk-heading-s">{{ "professions.admin.publish.blocked.heading" | t }}</h2>
+    <ul class="govuk-list govuk-list--bullet">
+      {%for blocker in publicationBlockers %}
+        {% if blocker.type === "incomplete-section" %}
+          {% set sectionName = (("professions.form.headings." + blocker.section) | t) %}
+          <li class="govuk-body-s">{{ "professions.admin.publish.blocked.missing" | t({section: sectionName}) }}</li>
+        {% elif blocker.type === "organisation-not-live" %}
```
Gweaton, line 9: We're not currently setting this, are we? Is for the next piece of work? If so, might be worth removing this section for now until we pick up that next task.

src/professions/admin/legislation.controller.ts
```diff
@@ -65,16 +61,13 @@ export class LegislationController {
...
     );
   }
 
   @Post('/:professionId/versions/:versionId/legislation')
   @Permissions(UserPermission.CreateProfession, UserPermission.EditProfession)
-  @BackLink((request: Request) =>
-    request.body.change === 'true'
-      ? '/admin/professions/:professionId/versions/:versionId/check-your-answers'
-      : '/admin/professions/:professionId/versions/:versionId/qualifications/edit',
+  @BackLink(
```
Gweaton, line 69: How come back links aren't going back to the CYA page now? I'd understand if we were getting rid of the order and they were all going back to CYA, but now they're always going back to the previous stage in the journey, even if you're editing an existing Profession - am I reading this right?

Gweaton: Either way, would be good to add a bit more to the commit message here, as this change feels a bit unrelated from the top level information change, so it's clearer for future readers why the changes were made here when looking at git history.

Gweaton (commented):
This is great work, I think it's a solid approach. Have just got one issue re: back links that might be a blocker, but I may just be misunderstanding the changes we're making here.

src/professions/admin/legislation.controller.ts
bryangaledxw: Gah, my bad, that's meant to be *always* going back to CYA, cheers for spotting

views/shared/_publication-blockers.njk
bryangaledxw: Ah, yep, agree

Gweaton (approved):
Nice 🏆

PR #478, opened 2022-03-22
(2263) Show alternate text for 2nd legislation for admins

Gweaton:
# Changes in this PR

Displays optional label text to admins for the second legislation when viewing a profession. This was based on UAT feedback. We don't show empty fields to members of the public, but it makes sense to show `(optional)` text on the field
here so as not to confuse admin users.

### Before

image

### After

image

### DIFF
--- a/views/professions/_profession-details.njk
+++ b/views/professions/_profession-details.njk
@@ -187,12 +187,15 @@
 <h2 class="govuk-heading-m rpr-details__section-title">{{ "professions.show.legislation.heading" | t}}</h2>
 
 {% for legislation in ((profession.legislations | pad(2)) if showEmptyProfessionDetails else profession.legislations) %}
+  {% set legislationLocalisationId = "professions.show.legislation.secondNationalLegislation" if (showEmptyProfessionDetails and loop.index > 1) else "professions.show.legislation.nationalLegislation" %}
+  {% set legislationLinkLocalisationId = "professions.show.legislation.secondLink" if (showEmptyProfessionDetails and loop.index > 1) else "professions.show.legislation.legislationLink" %}
+
   {{ govukSummaryList({
     classes: 'govuk-summary-list--no-border',
     rows: [
       {
         key: {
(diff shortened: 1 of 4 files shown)


views/professions/_profession-details.njk
```diff
@@ -187,20 +187,23 @@
 <h2 class="govuk-heading-m rpr-details__section-title">{{ "professions.show.legislation.heading" | t}}</h2>
 
 {% for legislation in ((profession.legislations | pad(2)) if showEmptyProfessionDetails else profession.legislations) %}
+  {% set legislationLocalisationId = "professions.form.label.legislation.secondNationalLegislation" if (showEmptyProfessionDetails and loop.index > 1) else "professions.show.legislation.nationalLegislation" %}
+  {% set legislationLinkLocalisationId = "professions.form.label.legislation.secondLink" if (showEmptyProfessionDetails and loop.index > 1) else "professions.show.legislation.legislationLink" %}
```
bryangaledxw, line 191: I don't think this is quite right - we're using `professions.show.foo.bar` on the details page and `professions.form.label.foo.bar` on the CYA page, the big difference being the details page doesn't mark things as "(Optional)". I don't think we want to mix-and-match the two on the same page

I think we want new strings, `professions.show.legislation.secondNationalLegislation` and `professions.show.legislation.secondLegislationLink`

Gweaton: So these would just be the same text without `(optional)`? I'm not against it, I thought showing "optional" might be clearer to users that they don't need to fill it in, but I suppose we don't show that with other pages, so it's probably better to be consistent.

bryangaledxw: Basically. The "optional" only really makes sense for anyone who's filling in details, and this would be show on the public-facing side

Gweaton: Have just updated it (and the screenshot in this PR) - is that what you were looking for?

bryangaledxw (approved):

PR #481, opened 2022-03-23
(2282) Display search results in alphabetical order

Gweaton:
# Changes in this PR

Ensures we're displaying professions and organisations in alphabetical order whenever they're shown in lists.

We were initially doing this in most places, but forgot to make sure we were doing this when we switched to DB-level searching rather than code-level - this wasn't caught by our e2e tests due to our reliance on the `sort` function in tests, which actually orders an array in-place, and was always returning true for our assertions.

### Before

image

image

image

image

#### Users

image

### After

image

image

image

image

#### Users

image

### DIFF
--- a/src/organisations/organisation-versions.service.ts
+++ b/src/organisations/organisation-versions.service.ts
@@ -90,12 +90,11 @@ export class OrganisationVersionsService {
   }
 
   async searchLive(filter: FilterInput): Promise<Organisation[]> {
-    const query = this.versionsWithJoins().where(
-      'organisationVersion.status = :status',
-      {
+    const query = this.versionsWithJoins()
+      .orderBy('organisation.name')
+      .where('organisationVersion.status = :status', {
         status: OrganisationVersionStatus.Live,
-      },
-    );
+      });
 
     return await this.filter(query, filter);
   }
--- a/src/organisations/organisation-versions.service.ts
+++ b/src/organisations/organisation-versions.service.ts
@@ -118,7 +118,11 @@ export class OrganisationVersionsService {
 
   async searchWithLatestVersion(filter: FilterInput): Promise<Organisation[]> {
     const query = this.versionsWithJoins()
-      .distinctOn(['organisationVersion.organisation', 'professions.id'])
+      .distinctOn([
+        'organisationVersion.organisation',
+        'organisation.name',
+        'professions.id',
+      ])
       .where(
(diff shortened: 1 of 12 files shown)


src/professions/profession-versions.service.ts
```diff
@@ -235,7 +232,7 @@ export class ProfessionVersionsService {
     organisation: Organisation,
   ): Promise<Profession[]> {
     const versions = await this.versionsWithJoins()
-      .distinctOn(['professionVersion.profession'])
+      .distinctOn(['profession.name'])
```
bryangaledxw, line 235: I assume this is needed but are we *definitely* sure this doesn't cause a change in behaviour? Say we have two professions with the same name, does one get eliminated?

Same for the similar change below

Am I remembering right that the fields in `distinctOn` need to be a subset of the fields in `orderBy`? My intuition is that we want to keep `.distinctOn(['professionVersion.profession'])` and then have

```
.orderBy(
        'professionVersion.profession, professionVersion.created_at', 'professionVersion.profession',
        'DESC',
      )
```

src/organisations/organisation-versions.service.ts
```diff
@@ -118,7 +118,7 @@ export class OrganisationVersionsService {
 
   async searchWithLatestVersion(filter: FilterInput): Promise<Organisation[]> {
     const query = this.versionsWithJoins()
-      .distinctOn(['organisationVersion.organisation', 'professions.id'])
+      .distinctOn(['organisation.name', 'professions.id'])
```
bryangaledxw, line 121: Same comment as for professions

bryangaledxw (approved):
Looks good but I'm definitely wary about the changes in `distinctOn`.

I'm fine if not, but is it worth doing ordering of professions listed on an organisation page as part of this PR?

src/professions/profession-versions.service.ts
Gweaton: Very well-spotted, this wasn't right at all. I've had to tweak your suggestion slightly - we have to add `profession.name` to the `distinctOn` array so it can be the first thing we `orderBy` (otherwise the results aren't sorted by `Profession(id/hash??)`). 

I added an Organisation with the same name to the test seeds while testing this (removed before committing) and it showed both, so we can be confident we're not removing entities with duplicate names here.

bryangaledxw (approved):

PR #493, opened 2022-03-25
(2268) Remove the external authentication step once beis confirm

bryangaledxw:
# Changes in this PR

- Remove "double authentication" step from internal pages

We some non-`/admin` pages unprotected where it makes sense. `/callback` and `/logout` need to remain unprotected, and it seemed right to leave pages linked to from the footer of the admin pages unprotected


Gweaton (approved):
Looks good, thanks for sorting - will test this out on staging

PR #508, opened 2022-03-28
(2175) Format phone numbers correctly

bryangaledxw:
# Changes in this PR

* Telephone numbers are validated on input, and formatted as international numbers on display

Guidelines referenced in commits are here - https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#telephone-numbers - and here - https://design-system.service.gov.uk/patterns/telephone-numbers/

### Before
localhost_3000_regulatory-authorities_council-of-registered-gas-installers (1)

### After
localhost_3000_regulatory-authorities_council-of-registered-gas-installers


src/organisations/admin/dto/organisation.dto.ts
```diff
@@ -39,7 +39,6 @@ export class OrganisationDto {
     message: 'organisations.admin.form.errors.phone.empty',
   })
   telephone: string;
-  fax?: string;
```
Gweaton, line 42: ah yeah, have been meaning to do this for a while! Should we remove it from the `Organisation.entity` class here too? It's not actually stored in the db, I don't think. We can also remove it from the seeds.

src/i18n/en/organisations.json
```diff
@@ -101,7 +101,8 @@
           "empty": "Enter the address for the regulatory authority"
         },
         "phone": {
-          "empty": "Enter the phone number for the regulatory authority"
+          "empty": "Enter the UK telephone number for the regulatory authority",
+          "invalid": "Enter the UK telephone number for the regulatory authority, like 01632 960 936"
```
Gweaton, line 105: Do you think we should show this as hint text before an error?

Gweaton: And maybe specify that users shouldn't include the country code if we're going down that route?

src/helpers/validate-telephone.helper.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+  it('returns false if the number contains characters that are neither spaces nor numerals', () => {
+    expect(validateTelephone('(020) 7215 5000')).toEqual(false);
+  });
+
+  it('returns false if the number is not a valid UK number', () => {
+    expect(validateTelephone('555 555')).toEqual(false);
+  });
+
+  it('returns false if the number contains a country code', () => {
+    expect(validateTelephone('00 44 20 7215 5000')).toEqual(false);
```
Gweaton, line 14: I'm not sure about this one - is the `google-libphonenumber` library able to detect if it starts with a country code or not? I wonder if we should allow users to enter a country code here.

views/professions/_profession-details.njk
```diff
@@ -83,7 +83,7 @@
             text: ("professions.show.bodies.phoneNumber" | t)
           },
           value: {
-            text: organisation.telephone
+            text: organisation.telephone | telephone
```
Gweaton, line 86: Do all organisations have telephone numbers here? Just thinking about data on prod being in draft but still being assigned to a Profession - will this break the page if `organisation.telephone` is `null`? Might be safer to add an `or ""` here (or handle this in the nunjucks formatter, if we're not already).

Gweaton (approved):
Thanks for taking care of this, a couple of things, but nothing too serious.

views/professions/_profession-details.njk
bryangaledxw: This follows the standard of the other formatters of returning the empty string for false-y input

src/organisations/admin/dto/organisation.dto.ts
bryangaledxw: Have removed entirely

src/i18n/en/organisations.json
bryangaledxw: Added hint text, and made it explicit that a country code shouldn't be included

src/helpers/validate-telephone.helper.spec.ts
Gweaton: As discussed, we'll leave this for now.

Gweaton (approved):

PR #514, opened 2022-03-29
(2267) Add overseas routes to recognition fields

Gweaton:
# Changes in this PR

Re-adds summary of recognition routes and link to more information for professionals from other countries, which were removed a couple of weeks ago while we figured out the best way to collect and present this information to users. Also adds mandatory radio buttons for selecting the route itself.

#### Qualification page

image

#### Check your answers

image

#### Viewing a profession

image

#### Qualification page

image

#### Check your answers page

image

#### Viewing a profession

image

### DIFF
--- a/src/i18n/en/professions.json
+++ b/src/i18n/en/professions.json
@@ -75,7 +75,7 @@
         "ukRecognitionUrl": "More about recognition within the UK (optional)",
         "otherCountriesRecognition": {
           "routes": {
-            "label": "What routes exist",
+            "label": "Routes to recognition for professionals outside the UK",
             "all": "Routes for professionals in all countries",
             "some": "Routes for professionals in some countries",
             "none": "No routes exist - all professionals must re-qualify"
@@ -154,6 +154,9 @@
         "ukRecognitionUrl": {
           "invalid": "Please enter a valid website address"
         },
+        "otherCountriesRecognitionRoutes": {
+          "empty": "Select what recognition routes exist for professionals qualified in another country"
+        },
         "otherCountriesRecognitionUrl": {
           "invalid": "Please enter a valid website address"
         }
@@ -203,6 +206,12 @@
       "ukRecognition": "Routes to recognition within the UK",
       "ukRecognitionUrl": "More about recognition within the UK",
       "otherCountriesRecognition": {
+        "routes": {
+          "label": "Recognition for overseas professionals",
+          "all": "All - there are routes to recognition for professionals from all countries",
+          "some": "Some - qualifications from some countries may be recognised, but professionals from other countries will have to re-qualify",
+          "none": "None - all professionals must re-qualify"
+        },
         "summary": "Summary of routes",
         "url": "More about recognition for professionals outside the UK"
       }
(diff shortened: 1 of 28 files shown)


src/i18n/en/professions.json
```diff
@@ -202,6 +205,12 @@
       "ukRecognition": "Routes to recognition within the UK",
       "ukRecognitionUrl": "More about recognition within the UK",
       "otherCountriesRecognition": {
+        "routes": {
+          "label": "Recognition for overseas professionals",
+          "all": "All - there are routes to recognition for professionals from all countries",
+          "some": "Some - qualifications from some countries may be recognised, but professionals from other countries will have to re-qualify",
+          "none": "None - all professionals must re-qualify"
```
bryangaledxw, line 212: I wonder if this would be nicer formatted like:
image
Though understand if this is unwanted scope creep!

bryangaledxw (approved):

src/i18n/en/professions.json
Gweaton: This text is actually for displaying on "show" pages, rather than in the radio buttons, as Al felt we could be more succinct with admin questions but should be more verbose for the public pages. I agree with the sentiment though!

src/professions/admin/dto/qualifications.dto.ts
```diff
@@ -26,4 +26,14 @@ export class QualificationsDto {
   @Transform(({ value }) => preprocessUrl(value))
   @ValidateIf((e) => e.ukRecognitionUrl)
   ukRecognitionUrl: string;
+
+  otherCountriesRecognitionSummary: string;
```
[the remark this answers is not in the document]
Gweaton: I think this is independent from the URL, or optional enough that it probably doesn't need enforcing, as it could be an alternative to providing a summary.

src/qualifications/presenters/qualification.presenter.ts
```diff
@@ -24,6 +24,13 @@ export default class QualificationPresenter {
     this.qualification && this.qualification.ukRecognitionUrl,
   );
 
+  readonly otherCountriesRecognitionSummary =
+    this.qualification && this.qualification.otherCountriesRecognitionSummary;
```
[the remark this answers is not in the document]
Gweaton: Good point. Have added this in, and made the UK recognition text input the same (a multiline textarea).

views/admin/professions/check-your-answers.njk
```diff
@@ -307,6 +307,23 @@
                   ]
                 }
               } if not isUK else undefined,
+              {
+                key: {
+                  text: ("professions.form.label.qualifications.otherCountriesRecognition.routes.label" | t)
+                },
+                value: {
+                  text: (qualification.adminSelectedOtherCountriesRecognitionRoutes or "") | t
```
bryangaledxw, line 315: Can this be `(qualification.adminSelectedOtherCountriesRecognitionRoutes | t) if qualification.adminSelectedOtherCountriesRecognitionRoutes else ""`, so we're not knowingly passing an invalid ID into the translation filter?

bryangaledxw (approved):
Looks good, just one more thing I spotted

views/admin/professions/check-your-answers.njk
Gweaton: As discussed, have made the presenter method return the raw value, which we append to the localisation id here in the template, as nunjucks seems to evaluate everything in an `if` oneliner for some reason...

PR #529, opened 2022-03-30
(2272) Read and write to the new organisations - professions relation

pezholio:
This builds on #502 to move all the professions - organisation logic from the `organisation` and `additionalOrganisation` fields to the new `ProfessionToOrganisation` one to many entity and removes the old relations. 

I've had to dispese with my usual "all commits must be green approach" as I wanted to keep the commit history readable (hopefully I've done that!). Sorry there's a lot here, but it really needs to be shipped as one feature.

Once this is done, we can update the UI so up to five organisations can be added, and have a role applied to each.


src/professions/admin/top-level-information.controller.spec.ts
```diff
@@ -102,11 +102,16 @@ describe('TopLevelInformationController', () => {
...
-          organisation: organisation,
-          additionalOrganisation: additionalOrganisation,
-        });
+        const profession = professionFactory.build(
+          {
+            name: 'Example Profession',
+          },
+          {
+            transient: {
+              organisations: [organisation, additionalOrganisation],
```
Gweaton, line 111: Are these transient params doing anything here? Are they supposed to be from another commit?

src/professions/admin/presenters/organisations-presenter.ts
```diff
@@ -0,0 +1,21 @@
...
+    this.organisations = organisations || [];
+  }
+
+  public list(): string {
+    if (this.organisations.length === 0) {
+      return '';
+    }
+
+    const organisations = this.organisations
+      .map((organisation) => organisation.name)
```
Gweaton, line 16: Possibly me being overly paranoid here, but do we need to `escape` the name here, as we're outputting HTML?

src/testutils/factories/profession.ts
```diff
@@ -1,12 +1,20 @@
...
 import { RegulationType } from '../../professions/profession-version.entity';
+import { ProfessionToOrganisation } from '../../professions/profession-to-organisation.entity';
+
 import industryFactory from './industry';
 import organisationFactory from './organisation';
 import qualificationFactory from './qualification';
 import userFactory from './user';
 
-class ProfessionFactory extends Factory<Profession> {
+type ProfessionTransientParams = {
```
Gweaton, line 13: Ah, I think this is the commit it should belong to. I can see why it's a tricky one to get in order though...

Gweaton (approved):
Thanks for dealing with all this - an enormous effort. Have noticed there are 2 "Update changelog" commits here and a couple of tiny things.

As I'm sure you're aware, this was a very large PR so I found it a bit tricky to review. I hope I haven't missed anything - let's make sure we test this properly on staging before prod to make sure all's working as expected.

src/professions/admin/presenters/organisations-presenter.ts
pezholio: Yeah, good call - have done 👍

src/testutils/factories/profession.ts
pezholio: Yeah, it's a bit of a head wrecker alright! Ended up doing it all in one big WIP commit and then breaking it down after the fact. Spent too much time chasing down failing tests!

PR #530, opened 2022-03-30
(Fix) Bump nestjs-i18n to 8.3.2

Gweaton:
# Changes in this PR

Bumps `nestjs-i18n` to 8.3.2 after updating the index template to output html from the intro translation.

We were previously trying to translate an array in the template here, which led to strangely outputed characters when we updated`nestjs-i18n`.

We may want to convert this template to not use translated text at all, but for now, this allows us to update the version of `nestjs-i18n`.


pezholio (approved):
LGTM 👍

PR #531, opened 2022-03-30
(2197) When creating professions with all nations selected display UK on the public results

bryangaledxw:
# Changes in this PR

- Consistently display "United Kingdom" where we would otherwise display a list of all four UK nations

Looking at the original UAT spreadsheet this ticket came from, the confusion seemed to come from there being a "United Kingdom" radio option, that then didn't get displayed as "United Kingdom" consistently, so this PR resolves that

### Before

image

### After

image


Gweaton (approved):
Do you think we could write a Cypress test for this? We're mocking the `NationsListPresenter` in our unit tests, so would be good to verify the real thing is working in one of our "show" tests, for example.

PR #535, opened 2022-03-30
Multiple organisations permissions bugfix

bryangaledxw:
# Changes in this PR

- Fix a bug causing Professions belonging to multiple Organisations to be uneditable by all but central users

### DIFF
--- a/src/users/helpers/check-can-view-profession.spec.ts
+++ b/src/users/helpers/check-can-view-profession.spec.ts
@@ -1,39 +1,179 @@
 import { createDefaultMockRequest } from '../../testutils/factories/create-default-mock-request';
-import organisationFactory from '../../testutils/factories/organisation';
 import professionFactory from '../../testutils/factories/profession';
-import { checkCanViewOrganisation } from './check-can-view-organisation';
 import { checkCanViewProfession } from './check-can-view-profession';
-import { ProfessionToOrganisation } from './../../professions/profession-to-organisation.entity';
-
-jest.mock('./check-can-view-organisation');
+import userFactory from '../../testutils/factories/user';
+import * as getActingUserModule from './get-acting-user.helper';
+import * as getTierOneOrganisationsFromProfessionModule from '../../professions/helpers/get-tier-one-organisations-from-profession.helper';
+import organisationFactory from '../../testutils/factories/organisation';
+import { UnauthorizedException } from '@nestjs/common';
 
 describe('checkCanViewProfession', () => {
-  it('passes in a Profession with an organisation to call CheckCanViewOrganisation', () => {
-    const organisation1 = organisationFactory.build();
-    const organisation2 = organisationFactory.build();
+  describe('when called as a service owner user', () => {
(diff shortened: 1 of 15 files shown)


src/users/helpers/check-can-view-profession.spec.ts
```diff
@@ -1,39 +1,179 @@
 import { createDefaultMockRequest } from '../../testutils/factories/create-default-mock-request';
-import organisationFactory from '../../testutils/factories/organisation';
 import professionFactory from '../../testutils/factories/profession';
-import { checkCanViewOrganisation } from './check-can-view-organisation';
 import { checkCanViewProfession } from './check-can-view-profession';
-import { ProfessionToOrganisation } from './../../professions/profession-to-organisation.entity';
-
-jest.mock('./check-can-view-organisation');
+import userFactory from '../../testutils/factories/user';
+import * as getActingUserModule from './get-acting-user.helper';
```
Gweaton, line 5: Why do you need `import *` here? In other files we've been able to do `import { getActingUser } from './users/helpers/get-acting-user.helper';` - is this so we can use `jest.spyOn` rather than mocking the import with `jest.mock('[import-path]')`?

I don't know which one I prefer, it's just different!

src/users/helpers/check-can-view-profession.ts
```diff
@@ -1,13 +1,24 @@
...
-  for (const org of getOrganisationsFromProfession(profession)) {
-    checkCanViewOrganisation(request, org);
+  const actingUser = getActingUser(request);
+
+  if (actingUser.serviceOwner) {
+    return;
+  }
+
+  if (
+    !getTierOneOrganisationsFromProfession(profession).some(
```
Gweaton, line 18: Would you mind storing this in a more human readable variable so it's a bit clearer at a glance what this means? The `!` makes it a bit less readable for me! Something like `belongsToTierOneOrganisation`?

Gweaton (approved):
Looks good to me, just a couple of minor points

src/users/helpers/check-can-view-profession.spec.ts
bryangaledxw: Basically. I'm leaning towards `spyOn` where I can - it's more strongly typed (no casting to as jest.Mock needed), and you don't end up with paths going into `jest.mock` that Visual Studio doesn't know to update when stuff moves around

Gweaton: Yeah, I don't like having to mock at top level - hadn't considered the advantages of typing too. Happy to start writing new tests like this.

bryangaledxw: And the import *, is because you can't just do jest.spyOn(foo), it has to be jest.spyOn(parentObject, 'foo'), so we need to go one level up with our import

bryangaledxw: Sadly I've not been able to get `spyOn` to work for getting at a classes constructor

PR #536, opened 2022-03-31
(2297) Add and tidy up support documentation

Gweaton:
# Changes in this PR

- Adds first line support documentation, including more info on creating and managing users, as well as linking to relevant support tasks
- Adds more context and links to live environments from README, so it's clear where newcomers to the project need to go


pezholio (approved):
Lovely stuff 👍

PR #547, opened 2022-04-05
(2322) Nations are being searched with AND rather than OR

bryangaledxw:
# Changes in this PR

* Fix bug causing selecting more nations in our search filters to give fewer results


pezholio (approved):
LGTM 👍

PR #554, opened 2022-04-06
(2288) Show multiple regulators and roles on professions page

pezholio:
This alters the regulator show pages to show their role above their details, and their name (as well as a link to the regulator page) in the summary list below

### Before

image

### After

image

### DIFF
--- a/cypress/integration/organisations/search/search.spec.ts
+++ b/cypress/integration/organisations/search/search.spec.ts
@@ -252,10 +252,9 @@ function findMatchingOrganisations(filter: (profession: any) => boolean) {
 
     const organisationNames = matchingProfessions
       .map((profession) => [
-        profession.organisation,
-        profession.additionalOrganisation,
+        profession.organisations.map((organisation) => organisation.name),
       ])
-      .flat();
+      .flat(Infinity);
 
     cy.readFile('./seeds/test/organisations.json').then((organisations) => {
       const liveOrganisations = organisations.filter(
--- a/src/professions/admin/profession-versions.controller.spec.ts
+++ b/src/professions/admin/profession-versions.controller.spec.ts
@@ -4,6 +4,10 @@ import { Test, TestingModule } from '@nestjs/testing';
 import { Response } from 'express';
 import { I18nService } from 'nestjs-i18n';
 import { Organisation } from '../../organisations/organisation.entity';
+import {
+  ProfessionToOrganisation,
+  OrganisationRole,
+} from './../profession-to-organisation.entity';
 import QualificationPresenter from '../../qualifications/presenters/qualification.presenter';
 import { createMockI18nService } from '../../testutils/create-mock-i18n-service';
 import industryFactory from '../../testutils/factories/industry';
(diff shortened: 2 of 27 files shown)


src/professions/admin/profession-versions.controller.spec.ts
```diff
@@ -154,7 +157,16 @@ describe('ProfessionVersionsController', () => {
     describe('when the Profession is complete', () => {
       describe('when the Profession has a single Organisation', () => {
         it('should return populated template params', async () => {
-          const profession = professionFactory.build();
+          const organisation = organisationFactory.build();
+
+          const profession = professionFactory.build({
+            professionToOrganisations: [
```
Gweaton: I've been wondering if this is something we should create a new factory for to save us having to do `as ProfessionToOrganisation` but I don't know if it saves us all that many lines of code - it might just be a bit easier to read?

src/professions/professions.controller.ts
```diff
@@ -41,17 +40,9 @@ export class ProfessionsController {
...
-        const organisation = Organisation.withLatestLiveVersion(
-          professionToOrganisation.organisation,
-        );
-
-        return {
-          ...organisation,
-          role: professionToOrganisation.role,
-        };
-      },
+    const tierOneOrganisations = getGroupedTierOneOrganisationsFromProfession(
```
Gweaton, line 43: Really nice to have that old logic extracted into this helper.

views/professions/show.njk.spec.ts
```diff
@@ -0,0 +1,54 @@
+import { JSDOM } from 'jsdom';
```
Gweaton, line 1: Interesting - I've never written view specs in JS before. When do you think we should opt for these over e2e tests? Is it just when we have too much data for seeds, or is it to reduce the overall number of e2e tests which are fairly slow and expensive to maintain?

Just wondering about how we make this decision going forward

Gweaton (approved):
Well done - this looks like it was a challenging task! I think this looks really good - just have a few questions.

views/professions/show.njk.spec.ts
pezholio: Yeah, I think it's just to reduce the overload of having seeds for every single edge case. Also changing seeds often breaks other tests which we then have to fix. I do wonder if it might be prudent to have specific seeds for specific test cases, but I dunno.

cypress/integration/organisations/search/search.spec.ts
```diff
@@ -252,10 +252,11 @@ function findMatchingOrganisations(filter: (profession: any) => boolean) {
...
     const organisationNames = matchingProfessions
       .map((profession) => [
-        profession.organisation,
-        profession.additionalOrganisation,
+        profession.organisations.map((organisation) => organisation.name),
       ])
-      .flat();
+      .flat(Infinity);
+
+    console.log(organisationNames);
```
richpjames, line 260: should this log be here?

richpjames, line 260: should this log be here?

pezholio: Oooh, good catch! A relic of debugging somewhere

src/professions/admin/profession-versions.controller.spec.ts
pezholio: Yeah, that's exactly the conversation I had in my head, I dunno if it's worth it. We do have that `organisations` transient param, but that doesn't allow us to specify the roles, and I dunno if it's worth extending it either.

PR #555, opened 2022-04-06
(2305) Add placeholder organisation user and admin user dashboards

bryangaledxw:
# Changes in this PR

- Placeholder dashboard for viewing recognition decision data

This is very much a placeholder page to set the groundwork for future work. Notably this punts on differing columns for service owner/non-service owner users, and figuring out who should have permission to view this page (currently admins only)

### Before

n/a

### After

localhost_3000_admin_decisions

### DIFF
--- a/cypress/integration/admin/decisions/index.spec.ts
+++ b/cypress/integration/admin/decisions/index.spec.ts
@@ -1,0 +1,122 @@
+import { format } from 'date-fns';
+
+type SeedDecisionDataset = {
+  profession: string;
+  organisation: string;
+  year: number;
+  status: string;
+};
+
+describe('Listing decision datasets', () => {
+  context('When I am logged in as admin', () => {
+    beforeEach(() => {
+      cy.loginAuth0('admin');
+      cy.visitAndCheckAccessibility('/admin/decisions');
+    });
+
+    it('Lists all decision datasets', () => {
+      cy.readFile('./seeds/test/decision-datasets.json').then(
+        (datasets: SeedDecisionDataset[]) => {
+          const datasetsToShow = getDisplayedDatasets(datasets);
+
+          cy.translate('decisions.admin.dashboard.search.foundPlural', {
+            count: datasetsToShow.length,
+          }).then((foundText) => {
+            cy.get('body').should('contain', foundText);
+          });
+
+          datasetsToShow.forEach((dataset, index) => {
+            cy.get('tbody tr')
+              .eq(index)
+              .then(($row) => {
+                cy.wrap($row).should('contain', dataset.profession);
+                cy.wrap($row).should('contain', dataset.year.toString());
+
+                cy.get('[data-cy=changed-by-text]').should('not.exist');
+                cy.wrap($row).should(
+                  'contain',
+                  format(new Date(), 'd MMM yyyy'),
+                );
+
+                cy.translate(`app.status.${dataset.status}`).then((status) => {
+                  cy.wrap($row).should('contain', status);
+                });
+              });
+          });
+        },
+      );
+    });
+  });
+
+  context('When I am logged in as organisation admin', () => {
+    beforeEach(() => {
+      cy.loginAuth0('orgadmin');
+      cy.visitAndCheckAccessibility('/admin/decisions');
+    });
+
+    it('Lists decision datasets for my organisation', () => {
+      cy.readFile('./seeds/test/decision-datasets.json').then(
+        (datasets: SeedDecisionDataset[]) => {
+          const datasetsToShow = getDisplayedDatasets(datasets).filter(
+            (dataset) => dataset.organisation === 'Department for Education',
+          );
+
+          cy.translate('decisions.admin.dashboard.search.foundPlural', {
+            count: datasetsToShow.length,
+          }).then((foundText) => {
+            cy.get('body').should('contain', foundText);
+          });
+
+          datasetsToShow.forEach((dataset, index) => {
+            cy.get('tbody tr')
+              .eq(index)
+              .then(($row) => {
+                cy.wrap($row).should('contain', dataset.profession);
+                cy.wrap($row).should('contain', dataset.year.toString());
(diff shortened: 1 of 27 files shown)


src/i18n/en/decisions.json
```diff
@@ -0,0 +1,19 @@
+{
+  "admin": {
+    "heading": "Recognition decision data",
```
Gweaton, line 3: Can we do a bit more nesting of these here? A bit like we do for the `professions.json` translations - could we have something like:
```
  "headings": {
    "main": "Recognition decision data",
     // then I can put edit: "Add recognition decision data" etc.
  },
```

src/i18n/en/decisions.json
```diff
@@ -0,0 +1,19 @@
...
+    "addButtonLabel": "Add decision data",
+    "manageHeading": "Manage existing decision data",
+    "viewDetails": "View details <span class='govuk-visually-hidden'>about {year} {organisaition} {profession} decision data</span>",
+    "tableHeading": {
```
Gweaton, line 7: We might want to nest this somewhere as well, as we'll have the dashboard table, and the table for entering the decision data itself. Should we nest this under `"dashboard"` or something? I feel it makes sense to group by page as much as possible, to make it easier for less-developer type people to come in and update these in future.

src/i18n/en/decisions.json
Gweaton: Oh, one more thing, but it might just be my text editor, this file seems to be indented with 4 spaces (we've used 2 elsewhere), but let me know if this is just a problem on my side

src/i18n/en/decisions.json
bryangaledxw: I've nested everything under decisions.admin.dashboard now. This does push it further away from what we're doing with profession/org admin pages (we've still got professions.admin.heading, etc. there) but yeah, I think this is a good change

src/i18n/en/decisions.json
bryangaledxw: Hmmm. I'm definitely seeing it as two
image

src/decisions/admin/presenters/decision-datasets.presenter.ts
```diff
@@ -0,0 +1,60 @@
+import { I18nService } from 'nestjs-i18n';
+import { IndexTemplate } from './../interfaces/index-template.interface';
+import { ListEntryPresenter } from './list-entry.presenter';
+import { Organisation } from '../../../organisations/organisation.entity';
+import { Table } from '../../../common/interfaces/table';
+import { DecisionDataset } from '../../decision-dataset.entity';
+
+export class DecisionDatasetsPresenter {
+  constructor(
+    private readonly userOrganisaiton: Organisation | null,
```
Gweaton, line 10: typo in `organisation` here.

src/decisions/admin/presenters/decision-datasets.presenter.ts
```diff
@@ -0,0 +1,60 @@
...
+      ? this.i18nService.translate('app.beis')
+      : this.userOrganisaiton.name;
+
+    return {
+      organisation,
+      decisionDatasetsTable: await this.table(),
+    };
+  }
+
+  private async table(): Promise<Table> {
```
Gweaton, line 26: Does this need to be `async` now that `translate` doesn't return a promise? I know we have another `ListEntryPresenter` but we sort of force that to return a promise at the moment as we haven't updated it.

src/decisions/admin/presenters/list-entry.presenter.ts
```diff
@@ -0,0 +1,63 @@
...
+        },
+      },
+    )}</a>`;
+
+    const entries: { [K in Field]: TableCell } = {
+      profession: { text: this.dataset.profession.name },
+      year: { text: this.dataset.year.toString() },
+      lastModified: { text: formatDate(this.dataset.updated_at) },
+      status: {
+        html: await formatStatus(this.dataset.status, this.i18nService),
```
Gweaton, line 56: Ah, this is why we need the `async` - I'll add a development task to go in and tidy this up, as it's a bit of a bigger job than is necessary for this PR.

seeds/development/decision-datasets.json
```diff
@@ -60,5 +60,58 @@
...
+            }
+          }
+        ]
+      }
+    ]
+  },
+  {
+    "profession": "Secondary School Teacher in State maintained schools (England)",
+    "organisation": "Department for Education",
+    "year": 2023,
```
Gweaton, line 98: I wonder if we want to make this in the past, as in future users will only be able to add data for the current and previous years. Also for the test seed json.

cypress/integration/admin/decisions/index.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    });
+  });
+
+  context('When I am logged in as organisation admin', () => {
+    beforeEach(() => {
+      cy.loginAuth0('orgadmin');
+      cy.visitAndCheckAccessibility('/admin/decisions');
+    });
+
+    it.only('Lists decision datasets for my organisation', () => {
```
Gweaton, line 57: An `only` has snuck in here.

Gweaton (approved):
This looks good to me - just a couple of minor things before merging.

src/i18n/en/decisions.json
Gweaton: Strange, it's now showing 2, my VS Code must've been doing something weird.

src/decisions/admin/presenters/list-entry.presenter.ts
bryangaledxw: Yep! I was 50/50 about changing `formatStatus` as part of this PR, but seemed a big task given we want to get this in to unblock other work

seeds/development/decision-datasets.json
bryangaledxw: Yep, agree

PR #568, opened 2022-04-11
(2327) Ensure admin organisation view still renders when a Regulator is removed from a Profession

Gweaton:
# Changes in this PR

Add guard against `ProfessionToOrganisation`s with a `null` `Profession`.

There's currently an issue with the underlying data when setting a new Regulator on a Profession, where the ProfessionToOrganisation isn't removed from the Regulator, but its `Profession` is set to `null`. This was causing the view to break.

As a temporary fix to this issue, while we fix the underlying data, we're allowing the page to render without errors - we'll run a migration to fix the data once this is solved.


pezholio (approved):
LGTM 👍

PR #570, opened 2022-04-11
Delete orphaned rows when updating relations

pezholio:
We’ve had an issue where when changing an profession’s organisations, where, rather than deleting the whole relation, the profession was getting set to `null` and the relation was staying around. The resulted in some strange behaviour and some views not loading.

This added the TypeORM option `orphanedRowAction`,which deletes any rows that are left orphaned when saved.

I considered writing a feature spec for this, but given #568 guards against the specific instance of the issue that we were seeing, I didn’t want to remove that fix.


richpjames (approved):
Looks good. 
I agree re: not writing a spec for this. We are testing against reoccurence of the bug higher up so I think any spec specifically for this would just be testing TypeORM functionality.

Gweaton (approved):
Thanks for sorting this - do you think we want to write a migration for the messy data?

pezholio:
@Gweaton Good call - I think I'll do it in another PR to give folks a chance to review it though 👍

PR #571, opened 2022-04-11
Add a data migration to remove orphaned relations

pezholio:
As per #570 - we have now fixed the issue where some profession to organisation relations were left behind when changing a profession's organisaion. However, there are still some orphaned relations hanging around, so this PR add a data migration to clean them up. These migrations will need to be run when deployed to staging/prod.


Gweaton (approved):
Thanks for this - looks good to me 👍

PR #576, opened 2022-04-12
(2306) Decision data entry table (part 1)

bryangaledxw:
# Changes in this PR

* Placeholder page for editing decision data

The ability to entry new decision data will follow in part 2

### Before

N/A

### After

localhost_3000_admin_decisions_16a00b20-4494-43d7-a9ae-212dda26064c_d2ea5676-e996-455e-ab52-3c435f39e1d5_2019_edit


src/decisions/admin/decisions.controller.ts
```diff
@@ -78,6 +112,169 @@ export class DecisionsController {
...
+      const noAfterComps = editDto.noAfterComps?.[routeIndex] || [];
+
+      const countriesCount = countries.length;
+
+      for (
+        let countryIndex = 0;
+        countryIndex < Math.min(countriesCount, maxCountryCount);
+        countryIndex++
+      ) {
+        const country: DecisionRoute['countries'][0] = {
```
Gweaton, line 215: Should the `0` here be `countryIndex`?

src/decisions/admin/decisions.controller.ts
```diff
@@ -78,6 +112,169 @@ export class DecisionsController {
...
+
+    const organisation = await this.organisationsService.find(organisationId);
+
+    checkCanViewOrganisation(request, organisation);
+
+    const routes: DecisionRoute[] = [];
+
+    const routeCount = editDto.routes?.length || 0;
+
+    for (
```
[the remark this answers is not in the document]
Gweaton: Putting it outside the controller class would also help reduce some of the complexity of the controller tests.

src/decisions/admin/decisions.controller.ts
```diff
@@ -78,6 +112,169 @@ export class DecisionsController {
...
+        routes,
+      };
+
+      await this.decisionDatasetsService.save(newDataset);
+
+      response.redirect(
+        `/admin/decisions/${profession.id}/${organisation.id}/${year}`,
+      );
+    } else {
+      const tokens = action.split(':');
```
Gweaton, line 250: I was wondering about how you'd approach this, after digging into it a bit myself when Ming was asking about the designs. We discovered the `formaction` attribute that can be added to a `button` https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formaction, which would allow us to create separate routes for each action e.g. publish, save as draft, remove route and submit to them via the form instead of always posting to the same route as we do here. Would that work, or would it complicate things further?

For me, there's a lot going on here for one controller action, which is why I'm wondering if there's an alternative way of breaking this down a bit.

Might be worth a quick chat about this though, as it's a fair bit to take in!

Gweaton (commented):
This looks great, can't believe you got it done so quickly. Would be good to think about how we can split up that heavy controller action though - happy to chat about it or pair on it if that would help?

src/decisions/decision-dataset.entity.ts
```diff
@@ -44,12 +44,12 @@ export class DecisionDataset {
...
-  created_at: Date;
+  created_at?: Date;
 
   @UpdateDateColumn({
     type: 'timestamp',
     default: () => 'CURRENT_TIMESTAMP(6)',
     onUpdate: 'CURRENT_TIMESTAMP(6)',
   })
-  updated_at: Date;
+  updated_at?: Date;
```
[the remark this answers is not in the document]
bryangaledxw: Hmm. What I don't want to do is change:

```
const newDataset: DecisionDataset = {
        organisation,
        profession,
        user: getActingUser(request),
        year: parseInt(year),
        status:
          action === 'publish'
            ? DecisionDatasetStatus.Live
            : DecisionDatasetStatus.Draft,
        routes,
      }
```

to

```
const newDataset = {
        organisation,
        profession,
        user: getActingUser(request),
        year: parseInt(year),
        status:
          action === 'publish'
            ? DecisionDatasetStatus.Live
            : DecisionDatasetStatus.Draft,
        routes,
      } as DecisionDataset
```

and lose the enforcement that all the actually required fields are given.

What I can do though change `updated_at`/`created_at` back to non-optional, and do:

```
const newDataset: DecisionDataset = {
        organisation,
        profession,
        user: getActingUser(request),
        year: parseInt(year),
        status:
          action === 'publish'
            ? DecisionDatasetStatus.Live
            : DecisionDatasetStatus.Draft,
        routes,
       updated_at: undefined,
       created_at: undefined
      }
```

src/decisions/admin/decisions.controller.ts
bryangaledxw: Ah, no, this is getting the type rather than a value - the `[0]` is getting the element type of the `countries` array on the `DecisionRoute` type

bryangaledxw: So it resolves to the type

```
  {
    country: string;
    decisions: {
      yes: number | null;
      no: number | null;
      yesAfterComp: number | null;
      noAfterComp: number | null;
    };
  }
```

src/decisions/admin/decisions.controller.ts
bryangaledxw: Ah, I'd ruled this out because I didn't think I could use it with `govukButton`, but there's the `attribute` param. Yeah, think this is worth using

bryangaledxw: Hmm. I've actually flip flopped on this. I think when you hit publish/add route/remove country etc., and if you hit save/publish but there are validation errors, we want the URL the user sees to still be `decisions/[prof]/[org]/[year]/edit`, and the the way to do that is by having these on the same route

We could save the data to the database and do a redirect to `decisions/[prof]/[org]/[year]/edit`, but we're not versioning our decision datasets, so we might be setting something that's published to unconfirmed

bryangaledxw: It does make it a lot cleaner though to pull out a couple of helper functions

src/decisions/admin/decisions.controller.ts
```diff
@@ -78,6 +110,113 @@ export class DecisionsController {
...
+      professionId,
+    );
+
+    checkCanChangeProfession(request, profession);
+
+    const organisation = await this.organisationsService.find(organisationId);
+
+    checkCanViewOrganisation(request, organisation);
+
+    const routes = parseEditDtoDecisionRoutes(editDto);
```
Gweaton, line 184: This looks much better!

src/decisions/admin/helpers/parse-decision-value.helper.ts
```diff
@@ -0,0 +1,9 @@
+export function parseDecisionValue(value: string): number | null {
+  const result = parseInt(value, 10);
+
+  if (isNaN(result) || result < 0) {
```
Gweaton, line 4: I've just tested this out locally - I think this should probably submit 0 instead of an empty string. We can fix that as part of a separate PR though, if you think it's worth doing?

Gweaton (approved):

src/decisions/admin/helpers/parse-decision-value.helper.ts
bryangaledxw: The idea was to make a distinction between "0" and a value that hasn't been filled in yet. It feels dangerous to silently replace an invalid value with a valid one?

Gweaton: Probably something we can do a bit of further testing on if necessary - let's leave it as is for now

PR #584, opened 2022-04-14
Refactor `formatStatus` as it's no longer async

Gweaton:
# Changes in this PR

Removes the `Promise` return from `formatStatus` since the new version of `nestjs-i18n`now returns a string.

I've done this now, as I'm updating code that touches `formatStatus` when editing the decision data dashboards and this felt like a good thing to squeeze in now, so we don't have to do a big bang approach later on.


richpjames (approved):
Looks good

PR #586, opened 2022-04-14
(2307) Add Regulator column to decision data table for admin users

Gweaton:
# Changes in this PR

Adds "Regulator" column to the table on the decision data index page for Central Admin users.

#### Regulator user

image

#### Central admin user

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Hide edit and delete buttons for non Tier 1 users
 - Placeholder page for editing decision data
 - Placeholder page for adding decision data
+- Add Regulator column to decision data table for central admin users
 
 ## [release-016] - 2022-04-11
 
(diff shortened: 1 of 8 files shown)


richpjames (approved):
LGTM

bryangaledxw (commented):
Did you take a look at the professions `ListEntryPresenter`? Where we're using ` 'overview'` and `'single-organisation'` as a key into a list of lists of fields. We're also using `type OrganisationsPresenterView = 'overview' | 'single-organisation'` in `OrganisationsPresenter`.

The code looks good, but my feeling is it's worth bring this more in line with those

Gweaton:
> Did you take a look at the professions `ListEntryPresenter`? Where we're using ` 'overview'` and `'single-organisation'` as a key into a list of lists of fields. We're also using `type OrganisationsPresenterView = 'overview' | 'single-organisation'` in `OrganisationsPresenter`.
> 
> The code looks good, but my feeling is it's worth bring this more in line with those

Thanks for letting me know - turns out I hadn't looked in enough detail at those classes for this. I've updated the PR now and think it looks a lot better. Would you mind taking another look @bryangaledxw?

bryangaledxw (approved):
LGTM

PR #587, opened 2022-04-14
(2306) Add simple table view for decision data entry (part 2)

bryangaledxw:
# Changes in this PR

## Screenshots of UI changes

* Placeholder page for adding decision data

This also at least partially addresses (2340) "Validate the year selection"

### Before

N/A

### After

localhost_3000_admin_decisions_new

### DIFF
--- a/src/decisions/admin/dto/new.dto.ts
+++ b/src/decisions/admin/dto/new.dto.ts
@@ -1,0 +1,24 @@
+import { Transform } from 'class-transformer';
+import { IsNotEmpty, ValidateIf } from 'class-validator';
+import { parseBoolean } from '../../../helpers/parse-boolean.helper';
+
+export class NewDto {
+  @Transform(({ value }) => parseBoolean(value))
+  serviceOwner: boolean;
+
+  @IsNotEmpty({
+    message: 'decisions.admin.new.errors.profession.empty',
+  })
+  profession: string;
+
+  @ValidateIf((e) => e.serviceOwner)
+  @IsNotEmpty({
+    message: 'decisions.admin.new.errors.organisation.empty',
+  })
+  organisation: string;
+
+  @IsNotEmpty({
+    message: 'decisions.admin.new.errors.year.empty',
+  })
+  year: string;
+}
--- a/src/decisions/admin/presenters/years-select.presenter.ts
+++ b/src/decisions/admin/presenters/years-select.presenter.ts
@@ -1,0 +1,31 @@
+import { I18nService } from 'nestjs-i18n';
+import { SelectItemArgs } from '../../../common/interfaces/select-item-args.interface';
+
+export class YearsSelectPresenter {
+  constructor(
+    private readonly startYear: number,
+    private readonly endYear: number,
+    private readonly selectedYear: number | null,
+    private readonly i18nService: I18nService,
+  ) {}
+
+  selectArgs(): SelectItemArgs[] {
+    const options = [
+      {
+        text: this.i18nService.translate<string>('app.pleaseSelect'),
+        value: '',
+        selected: null,
+      },
+    ];
+
+    for (let year = this.startYear; year <= this.endYear; year++) {
+      options.push({
+        text: year.toString(),
+        value: year.toString(),
+        selected: this.selectedYear === year,
+      });
+    }
+
+    return options;
+  }
+}
--- a/src/i18n/en/app.json
+++ b/src/i18n/en/app.json
@@ -93,5 +93,6 @@
   "filters": {
     "filterButton": "Filter results",
     "clearAllButton": "Clear all filters"
-  }
+  },
+  "pleaseSelect": "--- Please Select ---"
 }
--- a/src/professions/admin/presenters/professions-select.presenter.spec.ts
+++ b/src/professions/admin/presenters/professions-select.presenter.spec.ts
@@ -1,0 +1,108 @@
+import { SelectItemArgs } from '../../../common/interfaces/select-item-args.interface';
+import { createMockI18nService } from '../../../testutils/create-mock-i18n-service';
+import professionFactory from '../../../testutils/factories/profession';
+import { translationOf } from '../../../testutils/translation-of';
+import { ProfessionsSelectPresenter } from './professions-select.presenter';
+
+describe('ProfessionsSelectPresenter', () => {
+  describe('selectArgs', () => {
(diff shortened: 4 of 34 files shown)


src/i18n/en/app.json
```diff
@@ -93,5 +93,6 @@
   "filters": {
     "filterButton": "Filter results",
     "clearAllButton": "Clear all filters"
-  }
+  },
+  "pleaseSelect": "--- Please Select ---"
```
Gweaton, line 97: 👍

src/professions/profession-versions.service.ts
```diff
@@ -223,6 +223,20 @@ export class ProfessionVersionsService {
       .getMany();
   }
 
+  async allLiveForOrganisation(
+    organisation: Organisation,
```
Gweaton, line 227: I think we should pass in an `id` here, rather than the whole organisation, as we've done with other service methods.

src/decisions/admin/presenters/years-select.presenter.ts
```diff
@@ -0,0 +1,34 @@
...
+        selected: null,
+      },
+    ];
+
+    for (let year = this.startYear; year <= this.endYear; year++) {
+      options.push({
+        text: year.toString(),
+        value: year.toString(),
+        selected:
+          typeof this.selectedYear === 'number'
```
Gweaton, line 26: Out of interest, why have you opted for `typeof === 'number'` over `this.selectedYear !== null`?

Gweaton: Could we also use `this.selectedYear || false` here?

src/professions/admin/presenters/professions-select.presenter.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+          [profession1, profession2, profession3],
+          null,
+          createMockI18nService(),
+        );
+
+        expect(presenter.selectArgs()).toEqual(expected);
+      });
+    });
+    describe('when given a non-null selected profession', () => {
+      it('returns a list of professions with no profession selected', () => {
```
Gweaton, line 59: should this be `with the profession selected`?

src/decisions/admin/helpers/get-decisions-end-year.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    describe('when it is the end of the year', () => {
+      it('returns the previous year', () => {
+        // Months are zero-indexed
+        jest.useFakeTimers().setSystemTime(new Date(2026, 11, 31));
+
+        expect(getDecisionsEndYear()).toEqual(2025);
+      });
+    });
+  });
```
Gweaton, line 28: Just wanted to double check you have the same understanding that using fake timers here doesn't leak across test files? I was wondering if we'd need to do `useRealTimers` in an `afterEach` here.

https://jestjs.io/docs/timer-mocks says:

> While you can call jest.useFakeTimers() or jest.useRealTimers() from anywhere (top level, inside an it block, etc.), it is a global operation and will affect other tests within the same file. Additionally, you need to call jest.useFakeTimers() to reset internal counters before each test. If you plan to not use fake timers in all your tests, you will want to clean up manually, as otherwise the faked timers will leak across tests

It might be sensible to clean up in an `afterEach` anyway, so there's no state persisted between tests, e.g. if we add new tests to the file?

src/decisions/admin/dto/new.dto.ts
```diff
@@ -0,0 +1,24 @@
+import { Transform } from 'class-transformer';
+import { IsNotEmpty, ValidateIf } from 'class-validator';
+import { parseBoolean } from '../../../helpers/parse-boolean.helper';
+
+export class NewDto {
+  @Transform(({ value }) => parseBoolean(value))
+  serviceOwner: boolean;
```
Gweaton, line 7: What's this being used for? I'm reviewing by commit so it's hard to tell, but maybe I'll find out in a minute!

Gweaton (approved):
Looks good to me - have a few suggestions/questions.

src/decisions/admin/presenters/years-select.presenter.ts
bryangaledxw: Ah, I was sorta copying from `RegulatedAuthoritiesSelectPresenter` without considering why it was doing what it was doing. But I used `typeof this.selectedYear === 'number'` instead of just `this.selectedYear`, because as far as the presenter knows, the falsey value `0` is a valid year.

**_But_**, the who thing can just be `this.selectedYear === year`, as unlike with `RegulatedAuthoritiesSelectPresenter`, there's no attempting to resolve null to avoid

Thanks for spotting some excessively round-a-bout nonsense! 😛

src/professions/profession-versions.service.ts
bryangaledxw: We've got `allWithLatestVersionForOrganisation` and `latestVersion` in this service which also take an object but could take an ID string. My reasoning here was to give more type safety than just a string, and in practice, we're using this where we already have our hands on the organisation object

src/decisions/admin/dto/new.dto.ts
bryangaledxw: Ah, this is literally only used in `@ValidateIf((e) => e.serviceOwner)` on the `organisation` field. For non-service owner users, we want to allow this field to be blank, as we just use the acting user's organisation. But the only way to get at whether or not a user is a service owner within this way of doing validation, is to pass it from the page and into the DTO. That's why we have `<input type="hidden" name="serviceOwner" value="{{ user.serviceOwner }}" />` in `new.njk`

Obviously this can be hacked, but this is only the "display nice error messages to nice users" validation, not the "protect against malicious users" validation

src/decisions/admin/helpers/get-decisions-end-year.spec.ts
bryangaledxw: Yep, agree

src/professions/profession-versions.service.ts
Gweaton: That's fair, I'll defer to you on this

src/decisions/admin/dto/new.dto.ts
Gweaton: Ah I see - that makes sense, thanks for clarifying.

Gweaton (approved):

PR #595, opened 2022-04-19
2331 allow up to 25 regulators to be selected on a profession

richpjames:
# Changes in this PR
This change allows a profession to have up to 25 regulators. 
Whilst updating the clientside Javascript I also fixed a bug where the 'Add another regulator' button still showed even though the limit had been reached.
The clientside Javascript is responsible for showing and hiding each regulator as the 'Add another regulator' or 'Remove regulator' button is pressed. For users that have Javascript disabled this means that they will presented with a full list of 25 regulators that they can add details for when they load the page. This was judged to be an edge case so we are ok to have a stylistically degraded experience for users without javascript.

### Before
This screenshot erroneously shows the 'Add another regulator' button even though the limit is 5 due to a bug
image

### After
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [unreleased]
 
+### Changed
+
+- Allow Professions to have up to 25 regulators (previously 5)
+
 ### Added
 
 - Hide edit and delete buttons for non Tier 1 users
--- a/assets/javascripts/regulatory-authority-select.js
+++ b/assets/javascripts/regulatory-authority-select.js
@@ -10,18 +10,19 @@ function showRegulator(event) {
 
   const latestRegulator = items[0];
 
-  showElement(latestRegulator);
+  if (latestRegulator) {
+    showElement(latestRegulator);
 
-  const regulatoryAuthoritySelect = latestRegulator.querySelector(
-    'select.organisation',
-  );
-  const roleSelect = latestRegulator.querySelector('select.role');
+    const regulatoryAuthoritySelect = latestRegulator.querySelector(
+      'select.organisation',
+    );
+    const roleSelect = latestRegulator.querySelector('select.role');
 
-  regulatoryAuthoritySelect.removeAttribute('tabindex');
-  roleSelect.removeAttribute('tabindex');
+    regulatoryAuthoritySelect.removeAttribute('tabindex');
+    roleSelect.removeAttribute('tabindex');
 
-  addRemoveButton(latestRegulator);
-
+    addRemoveButton(latestRegulator);
+  }
   // If there are no more hidden containers left, hide the button.
   if (
     document.querySelectorAll(hiddenRegulatoryAuthorityContainerSelector)
--- a/assets/javascripts/regulatory-authority-select.js
+++ b/assets/javascripts/regulatory-authority-select.js
@@ -7,6 +7,7 @@ function showRegulator(event) {
   const items = document.querySelectorAll(
     hiddenRegulatoryAuthorityContainerSelector,
   );
+  const addRegulatorButton = document.querySelector('#add-regulator');
 
   const latestRegulator = items[0];
 
@@ -28,7 +29,8 @@ function showRegulator(event) {
     document.querySelectorAll(hiddenRegulatoryAuthorityContainerSelector)
       .length == 0
   ) {
-    hideElement(addButton);
+    addRegulatorButton.classList.add('rpr-hidden-element');
+    addRegulatorButton.setAttribute('aria-hidden', true);
   }
 
   event.preventDefault();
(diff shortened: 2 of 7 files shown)


src/professions/admin/organisations.controller.spec.ts
```diff
@@ -89,13 +89,10 @@ describe('OrganisationsController', () => {
...
           expect.objectContaining({
-            selectArgsArray: [
-              authoritiesAndRoles,
-              authoritiesAndRoles,
-              authoritiesAndRoles,
-              authoritiesAndRoles,
-              authoritiesAndRoles,
-            ],
+            selectArgsArray: Array(25)
+              .fill(undefined)
```
richpjames, line 93: I decided to `fill` the array with `undefined`s because if you `fill` an array with a non-primitive value then all the references point to the same value instead of 25 distinct objects as you might expect which could cause issues in the future. 
This does sacrifice readability however (although hopefully it is clear what is going on). I am interested to hear if anyone has a better way of doing this.

assets/javascripts/regulatory-authority-select.js
```diff
@@ -7,27 +7,30 @@ function showRegulator(event) {
   const items = document.querySelectorAll(
     hiddenRegulatoryAuthorityContainerSelector,
   );
+  const addRegulatorButton = document.querySelector('#add-regulator');
 
   const latestRegulator = items[0];
 
-  showElement(latestRegulator);
+  if (latestRegulator) {
```
richpjames, line 14: Accessing `latestRegulator` even when it didn't exist was causing errors in the console so I thought it is better to check that it exists.

cypress/integration/admin/professions/new.spec.ts
```diff
@@ -420,5 +420,40 @@ describe('Adding a new profession', () => {
         format(new Date(), 'd MMM yyyy'),
       );
     });
+    it("hides 'Add a Regulator' button when the maximum number of regulators (25) is reached ", () => {
+      cy.visit('/admin/professions');
```
richpjames, line 424: It irks me slightly having to do this set up again so let me know if there is a better way. I don't think it is worth extracing to a command as it is not something we want to be doing regularly

cypress/integration/admin/professions/new.spec.ts
```diff
@@ -420,5 +420,40 @@ describe('Adding a new profession', () => {
...
+
+      cy.clickSummaryListRowChangeLink(
+        'professions.form.label.organisations.name',
+      );
+
+      cy.get('select[id="regulatoryBodies_1"]').select(
+        'Department for Education',
+      );
+
+      for (let i = 0; i < 24; i++) {
```
richpjames, line 444: This is probably not great for test performance but it is the simplest way I thought of of testing that the button is hidden when we reach the maximum number of regulators. I thought it was important to test given that this functionality broke before.

cypress/integration/admin/professions/new.spec.ts
Gweaton: I think this is fine - we sometimes do this in a `beforeEach` block, but it's good to have all the context here in one test.

assets/stylesheets/application.scss
```diff
@@ -10,6 +10,10 @@ $small-screen: 768px;
   padding-top: 10px;
 }
 
+.rpr-hide-element {
```
Gweaton, line 13: can we call this `rpr-hidden-element`?

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [unreleased]
 
+- Allow Professions to have up to 25 regulators (previously 5)
```
Gweaton, line 9: This should be nested under a `### Changed` block (see how we've done it below)

src/professions/admin/organisations.controller.spec.ts
Gweaton: Does `Array.from({ length:25 },()=> (...authoritiesAndRoles))` work?

cypress/integration/admin/professions/new.spec.ts
Gweaton: I think this is fine!

Gweaton (approved):
Looks good, thanks for this - just a couple of minor things before merging.

assets/javascripts/regulatory-authority-select.js
```diff
@@ -69,14 +71,18 @@ function initialize(items) {
...
-
-  hideElement(container);
+  const addRegulatorButton = document.querySelector('#add-regulator');
 
   container.querySelector('select.organisation').value = '';
   container.querySelector('select.role').value = '';
 
   button.remove();
-
+  container.classList.add('rpr-hide-element');
```
Gweaton, line 80: This should be `rpr-hidden-element` now, right?

PR #600, opened 2022-04-19
(2336) Fix validation of year/profession/organisation combination when editing decision data

bryangaledxw:
# Changes in this PR

- Decision data can be submitted for any regulatory authority that regulates a profession in any role


src/decisions/admin/helpers/check-can-change-dataset.helper.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+        expect(getActingUserSpy).toHaveBeenCalledWith(request);
+        expect(getOrganisationsFromProfessionSpy).toHaveBeenCalledWith(
+          profession,
+        );
+        expect(getDecisionsYearsRangeSpy).toHaveBeenCalled();
+      });
+    });
+
+    describe('when the Profession is part of the given Organisation, and the year is in the allowed range', () => {
+      it('throws a BadRequstException', () => {
```
richpjames, line 324: I think this test title may need changing as it looks like the function doesn't throw an error for this call.

richpjames (approved):
Looks good to me but it might be worth waiting for @Gweaton to have a look before merging

src/decisions/admin/helpers/get-decisions-years-range.ts
```diff
@@ -0,0 +1,3 @@
+export function getDecisionsYearsRange(): { start: number; end: number } {
+  return { start: 2020, end: new Date().getFullYear() - 1 };
```
Gweaton, line 2: For future readers of the codebase, it might make sense to store `2020` in a variable like `firstYearOfRecordedDecisionData` or something, so it's less of a magic number and clear why we've hardcoded that value in?

src/decisions/admin/helpers/check-can-change-dataset.helper.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+        expect(getActingUserSpy).toHaveBeenCalledWith(request);
+        expect(getOrganisationsFromProfessionSpy).toHaveBeenCalledWith(
+          profession,
+        );
+        expect(getDecisionsYearsRangeSpy).not.toHaveBeenCalled();
+      });
+    });
+
+    describe('when the given year is before the earliest allowed year', () => {
+      it('throws a BadRequstException', () => {
```
Gweaton, line 53: typo here in `BadRequestException`. Looks like this might have been copied and pasted too, so you'll wanna replace it in a few places in this file.

src/decisions/admin/helpers/check-can-change-dataset.helper.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+            'getOrganisationsFromProfession',
+          )
+          .mockReturnValue([otherOrganisation]);
+        const getDecisionsYearsRangeSpy = jest.spyOn(
+          getDecisionsYearsRangeModule,
+          'getDecisionsYearsRange',
+        );
+
+        expect(() => {
+          checkCanChangeDataset(request, profession, organisation, 2020);
```
Gweaton, line 41: I'm finding it a bit hard to parse which of the variables here is causing the error to be thrown - to make it easier to read at a glance here, and so you don't have to go and figure out what the spies are doing, do you think we could make the names a bit more explicit here like `professionFromOtherOrganisation` or something?

Gweaton (approved):
This is great. Just a few small comments, mostly about naming.

src/decisions/admin/decisions.controller.ts
```diff
@@ -93,7 +94,7 @@ export class DecisionsController {
   async show(
     @Param('professionId') professionId: string,
     @Param('organisationId') organisationId: string,
-    @Param('year') year: string,
+    @Param('year', ParseIntPipe) year: number,
```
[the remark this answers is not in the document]
bryangaledxw: IKR! There's a parallel universe where we're doing `@Param('profession', FindProfessionPipe) profession: Profession`

src/decisions/admin/helpers/check-can-change-dataset.helper.ts
```diff
@@ -0,0 +1,59 @@
...
+import { getActingUser } from '../../../users/helpers/get-acting-user.helper';
+import { User } from '../../../users/user.entity';
+import { getDecisionsYearsRange } from './get-decisions-years-range';
+
+export function checkCanChangeDataset(
+  request: RequestWithAppSession,
+  profession: Profession,
+  organisation: Organisation,
+  year: number,
+  exists: boolean,
```
Gweaton, line 15: can we call this something like `isExistingDataset` just to be a bit more explicit for my own brain?

PR #601, opened 2022-04-20
2264 archiving a regulator shouldn't archive their professions

richpjames:
# Changes in this PR

This PR prevents admins from archiving a organisations whilst it has professions attatched. 

We could have chosen to automatically remove professions from a regulator when it is being archived but the user may not expect this to happen so could have been confusing. It also may have meant for more complex logic in the controller which would have to be maintained.

Instead if a user attempts to archive an organisation with professions they are presented with a screen (screenshot below) showing them which professions should be removed from the organisation in order for them to archive it.

### Before

image

### After
image

### DIFF
--- a/cypress/integration/admin/organisations/archive.spec.ts
+++ b/cypress/integration/admin/organisations/archive.spec.ts
@@ -7,187 +7,286 @@ describe('Archiving organisations', () => {
       cy.visitInternalDashboard();
     });
 
-    it('Allows me to archive a draft organisation', () => {
-      cy.get('a').contains('Regulatory authorities').click();
-      cy.checkAccessibility();
+    describe('Draft organisations', () => {
+      it('Allows me to archive a draft organisation with no associated professions', () => {
+        cy.get('a').contains('Regulatory authorities').click();
+        cy.checkAccessibility();
 
-      cy.contains('Department for Education')
-        .parent('tr')
-        .within(() => {
-          cy.get('a').contains('View details').click();
-        });
-
-      cy.translate('app.status.draft').then((status) => {
-        cy.get('h2[data-status]').should('contain', status);
-      });
-
-      cy.translate('organisations.admin.button.archive').then(
-        (archiveButton) => {
-          cy.get('a').contains(archiveButton).click();
-        },
-      );
-
-      cy.checkAccessibility();
-
-      cy.translate('organisations.admin.archive.caption').then(
-        (archiveCaption) => {
-          cy.get('body').contains(archiveCaption);
-        },
-      );
-
-      cy.translate('organisations.admin.archive.heading', {
-        organisationName: 'Department for Education',
-      }).then((heading) => {
-        cy.contains(heading);
-      });
-
-      cy.translate('organisations.admin.button.archive').then((buttonText) => {
-        cy.get('button').contains(buttonText).click();
-      });
-
-      cy.checkAccessibility();
-
-      cy.translate('organisations.admin.archive.confirmation.heading').then(
-        (confirmation) => {
-          cy.get('html').should('contain', confirmation);
-        },
-      );
-
-      cy.get('[data-cy=actions]').should('not.exist');
-
-      cy.translate('app.status.archived').then((status) => {
-        cy.get('h2[data-status]').should('contain', status);
-      });
-      cy.get('[data-cy=changed-by-user-name]').should('contain', 'Registrar');
-      cy.get('[data-cy=changed-by-user-email]').should(
-        'contain',
-        'beis-rpr+registrar@dxw.com',
-      );
-      cy.get('[data-cy=last-modified]').should(
-        'contain',
-        format(new Date(), 'd MMM yyyy'),
-      );
-
-      cy.visitAndCheckAccessibility('/admin/organisations');
-
-      cy.get('tr')
-        .contains('Department for Education')
-        .then(($header) => {
-          const $row = $header.parent();
-
-          cy.translate(`app.status.archived`).then((status) => {
-            cy.wrap($row).should('contain', status);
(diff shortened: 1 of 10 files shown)


cypress/integration/admin/organisations/archive.spec.ts
```diff
@@ -7,187 +7,280 @@ describe('Archiving organisations', () => {
...
-        cy.get('h2[data-status]').should('contain', status);
-      });
-
-      cy.get('[data-cy=currently-published-version-text]').within(($h2) => {
-        cy.translate('organisations.admin.publicFacingLink.heading').then(
-          (publicFacingLinkHeading) => {
-            cy.wrap($h2).should('contain', publicFacingLinkHeading);
+        cy.translate('organisations.admin.archive.forbidden').then(
+          (message) => {
+            console.log(message);
```
Gweaton, line 130: There's a `console.log` here.

Gweaton: Is it worth asserting that you see the listed Professions here too?

src/organisations/admin/organisation-archive.controller.ts
```diff
@@ -69,6 +70,13 @@ export class OrganisationArchiveController {
         organisationId,
         versionId,
       );
+
+    const professions = version.organisation.professionToOrganisations?.find(
```
Gweaton, line 74: Can `professionToOrganisations` be undefined here? If so, can it in the `new` action above?

src/organisations/organisation-versions.service.ts
```diff
@@ -248,10 +246,7 @@ export class OrganisationVersionsService {
     return version;
   }
 
-  async archive(
-    version: OrganisationVersion,
-    user: User,
-  ): Promise<OrganisationVersion> {
+  async archive(version: OrganisationVersion): Promise<OrganisationVersion> {
```
Gweaton, line 249: Now that we're only updating one entity (not the `professionVersion` too), I think we can simplify this function and stop needing transactions, so the body would be:
```
version.status = OrganisationVersionStatus.Archived;
return this.repository.save(version);
```

The tests will need simplifying, I think they'll end up looking a lot like the `confirm` function's tests.

views/admin/organisations/archive/new.njk
```diff
@@ -13,17 +14,33 @@
 
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
-      <p class="govuk-body">{{ 'organisations.admin.archive.details' | t }}</p>
+      {% if canArchiveOrganisation %}
+        <p class="govuk-body">{{ 'organisations.admin.archive.details' | t }}</p>
```
Gweaton, line 18: We can remove this `<p>` tag now, alongside the `details` translated text, as it says "This will archive all the professions", which is no longer the case.

cypress/integration/admin/organisations/archive.spec.ts
richpjames: I think so

src/organisations/admin/organisation-archive.controller.spec.ts
```diff
@@ -194,5 +195,32 @@ describe('OrganisationArchiveController', () => {
 
       expect(checkCanViewOrganisation).toHaveBeenCalledWith(req, organisation);
     });
+    it('throws BadRequestException when request is made to archive organisation with professions', async () => {
```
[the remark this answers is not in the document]
richpjames: Of course

src/organisations/admin/organisation-archive.controller.ts
richpjames: According to it's type it can't but it's initialised as `undefined` in `OrganisationFactory`.
Should I change the `OrganisationFactory` so that the organisations are always initialised with a `professionToOrganisations` array? or leave this as it is so w

Gweaton: It looks like we tend to use the `?` operator when calling the `professionToOrganisations` in other places too - would you mind putting one in the `new` method above just to be safe? I think it might be a leftover from professions not being removed propely from regulators, but I'd like to investigate this.

richpjames: Yeah no problem, it's best to be safe

src/organisations/admin/organisation-archive.controller.ts
```diff
@@ -71,7 +71,7 @@ export class OrganisationArchiveController {
         versionId,
       );
 
-    const professions = version.organisation.professionToOrganisations?.find(
+    const professions = version.organisation?.professionToOrganisations?.find(
```
Gweaton, line 74: Sorry, I might not have been clear here - I meant adding a `?` after `professionToOrganisations?.map` on line 51, rather than the organisation here.

Gweaton (commented):
The new tests look good - just one thing about the `?` operator.

Also, the e2e tests will need fixing but you're aware of that!

src/organisations/organisation-versions.service.spec.ts
```diff
@@ -395,75 +397,51 @@ describe('OrganisationVersionsService', () => {
...
+          const organisation = organisationFactory.build();
+          const repoSpy = jest.spyOn(repo, 'findOne');
+
+          const result = await service.hasLiveVersion(organisation);
+
+          expect(result).toEqual(false);
+          expect(repoSpy).toHaveBeenCalledWith({
+            where: {
+              organisation: { id: organisation.id },
+              status: In([
```
richpjames, line 435: @Gweaton are these the correct statuses to be checking? I wasn't sure if it should be all three or just 'Live'

Gweaton (approved):
Sorry about the back and forth on this - have tested locally and all looks as expected. Good to hit merge!

PR #608, opened 2022-04-21
(2310) Add "bulk download" to admin/org dashboard

bryangaledxw:
# Changes in this PR

- Add bulk decision data download link

Our `DecisionsController` is becoming somewhat unwieldy now - I'm wondering if `new`/`newPost` and `edit`/`editPost` each belong in their own controller?

When we get to implementing filters, my thoughts are that the export route will take the filter selections via query params in the URL. When we generate the page following the user hitting the "Filter" button, the query params would be added onto the export link in `index.njk`

### Before
localhost_3000_admin_decisions_ (3)

### After
localhost_3000_admin_decisions_ (1)


src/decisions/admin/helpers/decisions-csv-writer.helper.ts
```diff
@@ -0,0 +1,72 @@
...
+    private readonly datasets: DecisionDataset[],
+    private readonly i18nService: I18nService,
+  ) {}
+
+  write(): void {
+    const stringifier = stringify();
+
+    this.response.set({
+      'Content-Type': 'text/csv',
+      'Content-Disposition': `attachment; filename="${this.filename}.csv"`,
```
Gweaton, line 21: I wonder if we want to put in the date of download into the filename here - can be something that comes as a separate PR if you agree though.

src/decisions/admin/decisions.controller.ts
```diff
@@ -85,6 +86,36 @@ export class DecisionsController {
     return this.createListEntries(request);
   }
 
+  @Get('export')
+  @Permissions(
+    UserPermission.UploadDecisionData,
+    UserPermission.DownloadDecisionData,
+    UserPermission.ViewDecisionData,
+  )
+  async export(
```
Gweaton, line 95: As you've said in the PR description about this controller getting a bit long, it could be an idea to move this into its own controller (`decisions-export.controller`?) with a `new` method. I think we'll end up splitting some things out of the controller fairly soon, so this might be a good point for that?

src/decisions/admin/helpers/decisions-csv-writer.helper.ts
```diff
@@ -0,0 +1,72 @@
...
+export class DecisionsCsvWriter {
+  constructor(
+    private readonly response: Response,
+    private readonly filename: string,
+    private readonly datasets: DecisionDataset[],
+    private readonly i18nService: I18nService,
+  ) {}
+
+  write(): void {
+    const stringifier = stringify();
```
Gweaton, line 17: I remember Stu advising on using streaming for creating the CSV file here, and I believe that's what [the CSV Stringify library recommends](https://csv.js.org/stringify/api/#stream-api):

> This is the recommended approach if you need a maximum of power. It ensures scalability by treating your data as a stream from the source to the destination.

Do you think it's worth using that API instead? I found the API docs for the library a bit hard to figure out, so couldn't see where the API you've used here came from, so correct me if you're already using it.

We used this on my previous project [here](https://github.com/ministryofjustice/hmpps-interventions-ui/blob/7ba5348a9d98b3ee4ab1b2f05e0d45fe74323f10/server/utils/serviceProviderCSVGenerator.ts#L36), if that's helpful at all?

Gweaton (commented):
This looks good - a couple of things, the main one re: which API to use on `stringify`

Gweaton (approved):

PR #609, opened 2022-04-25
(2348) Don't display duplicate organisations on public listing and when selecting an organisation for a profession

Gweaton:
# Changes in this PR

Adds a `distinctOn` clause to `OrganisationVersionsService` functions to filter out where we were displaying multiple versions on the public index of Regulators, as well as on the internal list of regulators selected for an Organisation.

### Before

image

### After

I can't actually replicate this locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Allow Professions to have up to 25 regulators (previously 5)
 - Show full list of countries when editing/adding decision data
 - Fix bug where publish button could still be clicked despite being disabled
+- Fix bug where duplicate regulators were being displayed publicly
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


richpjames (approved):
Looks good!

bryangaledxw (approved):
I'd double check that `organisation.name` is actually necessary - I _think_ it's redundant. If I've got it correct, if two things have a different name or are a different organisation, then they're distinct, but if the first is false the second is false.

Gweaton:
> I'd double check that `organisation.name` is actually necessary - I _think_ it's redundant. If I've got it correct, if two things have a different name or are a different organisation, then they're distinct, but if the first is false the second is false.

@bryangaledxw I thought the same, but you need to have `.name` in there as it's in `orderBy`, so if we want to sort them alphabetically we need it in there, otherwise we get: `SELECT DISTINCT ON expressions must match initial ORDER BY expressions`

PR #617, opened 2022-04-26
(2316) Tweaks to decision data publishing

Gweaton:
# Changes in this PR

- Adds "Publish to register" button to admin page for viewing Decision Data, only visible to Central Admin users
- Hides publish button from non-admin users
- Changes "edit" link to button when viewing decision data

#### Before

image

#### After

image

#### After - Publish to register visible on both edit and non-edit pages

image

image

image


views/admin/decisions/show.njk
```diff
@@ -12,6 +14,22 @@
     <div class="govuk-grid-column-one-quarter">
       <a class="govuk-link" href="/admin/decisions/{{profession.id}}/{{organisation.id}}/{{year}}/edit">{{ "decisions.admin.show.edit" | t }}</a>
     </div>
+
+    <nav role="navigation" data-cy="actions">
+      <ul class="govuk-list">
+        {% if 'publishDecisionData' in permissions %}
```
bryangaledxw: Do we want to hide this if the dataset is already published?

bryangaledxw: Do we want to hide this if the dataset is already published?

views/admin/decisions/edit.njk
```diff
@@ -177,11 +177,13 @@
           classes: "govuk-button--secondary"
         }) }}
 
-        {{ govukButton({
-          text: "decisions.admin.buttons.publish" | t,
-          name: 'action',
-          value: 'publish'
-        }) }}
+        {% if 'publishDecisionData' in permissions %}
```
bryangaledxw, line 180: I'd prefer not to leave this "half-enforced" - I think in the same commit, change the `update` method to also look at the user's permissions if the action is `publish`

src/decisions/admin/publication.controller.ts
```diff
@@ -0,0 +1,55 @@
...
+  async create(
+    @Param('professionId') professionId: string,
+    @Param('organisationId') organisationId: string,
+    @Param('year', ParseIntPipe) year: number,
+    @Req() request: RequestWithAppSession,
+    @Res() response: Response,
+  ) {
+    const user = getActingUser(request);
+
+    if (!user.serviceOwner) {
```
bryangaledxw, line 84: Do we want this redundancy, since we have `@Permissions(UserPermission.PublishDecisionData)`?

bryangaledxw (approved):
Looks good, I'm happy for this to go with the additional permission check in `edit.controller` - I worry that if we *just* hide the button, doing any actual enforcement will get forgotten

This still leaves us with the problem:
- User A looks at the data, is happy with it, clicks publish, goes to the confirm screen
- User B edits the data
- User A hits publish, and different data than what they thought they were publishing is published

Happy for this to stay unsolved for now, but I think it's worth capturing that we have this issue in a development task

views/admin/decisions/show.njk
Gweaton: Good point. I'll update this.

src/decisions/admin/publication.controller.ts
Gweaton: Have removed this

views/admin/decisions/edit.njk
Gweaton: Have done this in c7d26e71fa

Gweaton:
> This still leaves us with the problem:
> 
>     * User A looks at the data, is happy with it, clicks publish, goes to the confirm screen
> 
>     * User B edits the data
> 
>     * User A hits publish, and different data than what they thought they were publishing is published
> 
> 
> Happy for this to stay unsolved for now, but I think it's worth capturing that we have this issue in a development task

@bryangaledxw I'm hoping we can tackle it as part of the "submitted" task, but if we don't I'll add a task for it.

bryangaledxw (approved):

PR #622, opened 2022-04-26
(2360) Display confirmation flash message when publishing or saving decision data

Gweaton:
# Changes in this PR

Adds confirmation banner to pages after saving or publishing decision data. I've tweaked the content slightly from the designs to be a bit clearer.

### Save as draft

image

### Publication

image


richpjames (approved):
Looks good

PR #623, opened 2022-04-27
(2349) Implement decision data filers

bryangaledxw:
# Changes in this PR

- Filters on decision data dashboard

Note that I've extracted ticket 2366 to bring the decisions dashboard HTML more in-line with the prototype

### Before
localhost_3000_admin_decisions_ (3)

### After
localhost_3000_admin_decisions_ (1)

### DIFF
--- a/src/decisions/admin/presenters/decision-dataset-statuses-checkbox.presenter.ts
+++ b/src/decisions/admin/presenters/decision-dataset-statuses-checkbox.presenter.ts
@@ -1,0 +1,20 @@
+import { I18nService } from 'nestjs-i18n';
+import { CheckboxItems } from '../../../common/interfaces/checkbox-items.interface';
+import { DecisionDatasetStatus } from '../../decision-dataset.entity';
+
+export class DecisionDatasetStatusesCheckboxPresenter {
+  constructor(
+    private readonly checkedStatuses: DecisionDatasetStatus[],
+    private readonly i18nService: I18nService,
+  ) {}
+
+  checkboxItems(): CheckboxItems[] {
+    return Object.values(DecisionDatasetStatus)
+      .filter((status) => status !== DecisionDatasetStatus.Unconfirmed)
+      .map((status) => ({
+        text: this.i18nService.translate<string>(`app.status.${status}`),
+        value: status,
+        checked: this.checkedStatuses.includes(status),
+      }));
+  }
+}
--- a/src/decisions/decision-datasets.service.spec.ts
+++ b/src/decisions/decision-datasets.service.spec.ts
@@ -54,83 +54,472 @@ describe('DecisionDatasetsService', () => {
   });
 
   describe('all', () => {
-    it('returns all live and draft DecisionDatasets', async () => {
-      const datasets = decisionDatasetFactory.buildList(3);
+    describe('when an empty FilterInput is given', () => {
+      it('returns all live and draft DecisionDatasets', async () => {
+        const queryResultDatasets = decisionDatasetFactory.buildList(3);
 
-      const queryBuilder = createMock<SelectQueryBuilder<DecisionDataset>>({
-        leftJoinAndSelect: () => queryBuilder,
-        where: () => queryBuilder,
-        orderBy: () => queryBuilder,
-        getMany: async () => datasets,
+        const queryBuilder = createMock<SelectQueryBuilder<DecisionDataset>>({
+          leftJoinAndSelect: () => queryBuilder,
+          where: () => queryBuilder,
+          orderBy: () => queryBuilder,
+          getMany: async () => queryResultDatasets,
+        });
+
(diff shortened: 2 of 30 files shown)


src/decisions/decision-datasets.service.ts
```diff
@@ -77,4 +84,56 @@ export class DecisionDatasetsService {
...
+        statuses: filter.statuses,
+      });
+    }
+
+    return query;
+  }
+
+  private createLikeString(likeString: string): string {
+    // TypeORM doesn't escape this for us - see https://github.com/typeorm/typeorm/issues/5012
+    return `%${likeString.replace(/[\\%_]/g, (match) => `\\${match}`)}%`;
```
Gweaton, line 137: Oof, well done for sorting this.

src/decisions/admin/presenters/decision-dataset-statuses-checkbox.presenter.ts
```diff
@@ -0,0 +1,20 @@
+import { I18nService } from 'nestjs-i18n';
+import { CheckboxItems } from '../../../common/interfaces/checkbox-items.interface';
+import { DecisionDatasetStatus } from '../../decision-dataset.entity';
+
+export class DecisionDatasetStatusesCheckboxPresenter {
```
Gweaton, line 5: Do you think we could expand this in future to work with Org/Profession statuses? I was going to suggest we could make this a more generic `StatusesCheckboxPresenter`, but I suppose it might be tricky given the disparity between those statuses and the `DecisionDatasetStatuses`?

views/admin/decisions/index.njk
```diff
@@ -31,6 +31,17 @@
     </div>
   </div>
 
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+        {% if view !== 'overview' %}
```
Gweaton, line 36: I think this should be indented one less. Sorry, really wish we had a nunjucks linter, maybe we'll have to write our own.

Gweaton (approved):
Well done on this - the approach makes sense to me, I'm glad we're using the same pattern as the other internal dashboards.

Have just left a couple of minor comments re: tests.

src/decisions/admin/presenters/decision-dataset-statuses-checkbox.presenter.ts
bryangaledxw: Yeah, I did think maybe it'd be worth doing that - my feeling is leave it until we actually want that. I feel like make it generic when we're actually gonna use it as generic, then we have a better idea of how it'll work in practice

src/i18n/en/decisions.json
```diff
@@ -4,7 +4,32 @@
       "heading": "Recognition decision data",
       "addButtonLabel": "Add decision data",
       "manageHeading": "Manage existing decision data",
+      "showFilters": "Use the filters to choose the data sets to see and download",
```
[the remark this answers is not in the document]
bryangaledxw: Ah, this came from Ming's prototype. Have changed to "dataset" for consistency. Maybe this is something we'll have a better idea on after user research. Suggest I make a "iterate dashboard text" ticket and assign to Al?

src/decisions/decision-datasets.service.spec.ts
```diff
@@ -54,83 +54,472 @@ describe('DecisionDatasetsService', () => {
...
+          'organisation.name': 'ASC',
+          year: 'DESC',
+        });
+        expect(queryBuilder.getMany).toBeCalledWith();
+      });
+    });
+
+    describe('when a FilterInput with organisations is given', () => {
+      it('returns all live and draft DecisionDatasets for the given organisations', async () => {
+        const datasets = decisionDatasetFactory.buildList(3);
```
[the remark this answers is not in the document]
bryangaledxw: Does it help to rename this to something like `queryResultDatasets`?

I kinda feel putting the keywords/orgs on these obscures that no actual filtering is happening in our code, and what we're really testing is what gets called on the query builder, and that what comes out at the end is whatever the query builder gives us. `expect(result).toEqual(datasets);` is kinda the least important line in our test, the important bits are all the things that get called along the way.

I'm also already struggling with how long and repetitious some of our tests our and losing just being able to do `decisionDatasetFactory.buildList(3);` wouldn't help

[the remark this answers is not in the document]
Gweaton: Good points. I agree, the tests are getting a little unwieldy here. It's been hard to strike a balance between being self-contained and not repetitive, but I don't think right now is the time to do a big refactor of the codebase to improve that.

If you could rename the variable for now, that would help parse the tests a bit quicker. I think what made this a bit harder for me was the number and length of tests being added in one go, which is understandable due to the way we're writing tests at the moment.

PR #627, opened 2022-04-28
(2352) Allow org-level users to submit decision data

Gweaton:
# Changes in this PR

Adds a submit button and functionality to decision data for non-central admin users.

There's still some work to do around the order of statuses here, i.e. whether you should be able to publish a draft status as a central admin, whether you should be able to submit or edit an already-publishedstatus, but this is something we'll iterate on.

### To come in a separate PR:

- Add "submit" to page where users edit data - we'll likely want to refactor our Edit controller for this, and I don't want this PR to get too long to review.

### After

image

image

image


src/decisions/admin/submission.controller.ts
```diff
@@ -0,0 +1,57 @@
...
+  @Put('/:professionId/:organisationId/:year/submit')
+  @Permissions(UserPermission.SubmitDecisionData)
+  async create(
+    @Param('professionId') professionId: string,
+    @Param('organisationId') organisationId: string,
+    @Param('year', ParseIntPipe) year: number,
+    @Res() response: Response,
+    @Req() request: Request,
+  ) {
+    const dataset = await this.decisionDatasetsService.find(
```
bryangaledxw, line 75: Do we want `checkCanChangeDataset` here? `@Permissions(UserPermission.SubmitDecisionData)` alone isn't enough to check that someone is only submitting for their own organisation

bryangaledxw (approved):
LGTM

PR #628, opened 2022-04-28
(2350) Bulk download respects decision data filters

bryangaledxw:
# Changes in this PR

- Bulk CSV export now respects filters

An additional change I've included in this PR is the the exported CSV's filename now includes a timestamp. This makes me nervous from a test robustness PoV. Getting Cypress to inspect a wildcard filename (`decisions-*`) appears to be a Hard Problem, so we rely on both the test server, and the test runner agreeing on the date. Open to thoughts on whether this is a safe assumption or not

### DIFF
--- a/cypress/integration/admin/decisions/index.spec.ts
+++ b/cypress/integration/admin/decisions/index.spec.ts
@@ -336,7 +336,7 @@ function checkCsvDownload(
 
   const filename = path.join(
     Cypress.config('downloadsFolder'),
-    'decisions.csv',
+    `decisions-${format(new Date(), 'yyyyMMdd')}.csv`,
   );
 
   cy.readFile(filename).then((file) => {
--- a/cypress/integration/admin/decisions/index.spec.ts
+++ b/cypress/integration/admin/decisions/index.spec.ts
@@ -51,6 +51,8 @@ describe('Listing decision datasets', () => {
           });
         },
       );
+
+      checkCsvDownload(() => true);
     });
 
     it('I can filter by keyword', () => {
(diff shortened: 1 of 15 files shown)


src/decisions/admin/presenters/decision-datasets.presenter.ts
```diff
@@ -22,7 +22,9 @@ export class DecisionDatasetsPresenter {
     private readonly i18nService: I18nService,
   ) {}
 
-  present(view: DecisionDatasetsPresenterView): IndexTemplate {
+  present(
+    view: DecisionDatasetsPresenterView,
+  ): Omit<IndexTemplate, 'filterQuery'> {
```
Gweaton, line 27: I didn't know about this, I think it's quite a nice approach.

cypress/integration/admin/decisions/index.spec.ts
```diff
@@ -336,7 +336,7 @@ function checkCsvDownload(
 
   const filename = path.join(
     Cypress.config('downloadsFolder'),
-    'decisions.csv',
+    `decisions-${format(new Date(), 'yyyyMMdd')}.csv`,
```
Gweaton, line 339: I think it's safe to assume the test runner and the server agreeing on the date here, particularly as it's a day in the filename, rather than seconds/minutes etc. 

Alternatively, you could look into using [Cypress's clock](https://docs.cypress.io/api/commands/clock#Now) and see if that lets you hardcode a date in this test, but I don't think it's really necessary here.

Gweaton (approved):
This looks great, thanks for your hard work on it.

PR #630, opened 2022-04-29
2319 add sidebar to admin decision data page

richpjames:
# Changes in this PR
This PR adds a sidebar to the admin decision data page which contains:

- The status of the decision data (published/draft/submitted)
- The date that the data was last changed
- The user that last changed the data
- A publicly facing link for the data (currently intentionally blank until functionality has been developed)
- A button to edit the data

A button to download the dataset will be added in a subsequent PR

### Before
image

### After
image

### DIFF
--- a/cypress/integration/admin/decisions/show.spec.ts
+++ b/cypress/integration/admin/decisions/show.spec.ts
@@ -1,3 +1,5 @@
+import { format } from 'date-fns';
+
 describe('Showing a decision dataset', () => {
   context('When I am logged in as editor', () => {
     beforeEach(() => {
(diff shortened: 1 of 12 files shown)


src/decisions/admin/decisions.controller.spec.ts
```diff
@@ -433,11 +433,30 @@ describe('DecisionsController', () => {
         mockTables,
       );
 
+      Object.defineProperties(DecisionDatasetPresenter.prototype, {
+        changedBy: {
```
richpjames, line 437: This is a workaround for this bug: https://github.com/facebook/jest/issues/9675
TLDR: Jest's spyOn doesn't work for instance getters

richpjames: This means we can't assert that the presenter has been called with the expected values as we can't spy on the getters but this functionality is tested, at a higher level, in the e2e tests

src/helpers/format-status.helper.spec.ts
```diff
@@ -63,7 +63,7 @@ describe('formatStatus', () => {
     it('returns a yellow tag', () => {
       const i18nService = createMockI18nService();
 
-      const result = formatStatus(OrganisationVersionStatus.Draft, i18nService);
+      const result = formatStatus(DecisionDatasetStatus.Draft, i18nService);
```
Gweaton, line 66: I'm a bit unclear on the reason behind this change. The commit message suggests this is adding a new status, but weren't we already showing the status in the decision data table? I think we can probably remove this commit.

If you're keen on having a test for this status, it's probably an idea to either do it in its own test, or test a couple of variables in this test with different `Draft` statuses, but I don't think it's necessary here.

src/decisions/presenters/decision-dataset.presenter.spec.ts
```diff
@@ -180,6 +182,33 @@ describe('DecisionDatasetPresenter', () => {
...
+
+        const presenter = new DecisionDatasetPresenter(dataset, i18nService);
+
+        expect(presenter.changedBy).toEqual({
+          name: 'example',
+          email: 'test@test.com',
+        });
+      });
+
+      it("if the dataset doesn't have a user it returns their name and email address", () => {
```
Gweaton, line 203: Should this be `returns null`?

cypress/integration/admin/decisions/show.spec.ts
```diff
@@ -77,5 +79,92 @@ describe('Showing a decision dataset', () => {
...
+            .contains(publicFacingLinkButtonText)
+            .contains('a')
+            .click()
+            .url()
+            .should('contain', '');
+        },
+      );
+    });
+
+    it('I can publish the data by clicking the publish button', () => {
```
Gweaton, line 152: This is covered by the `publish.spec` now, so I think we can remove it from here.

views/admin/decisions/show.njk
```diff
@@ -26,7 +51,7 @@
               }}
             </li>
           {% endif %}
-          {% if 'submitDecisionData' in permissions and not (datasetStatus === 'submitted') %}
+          {% if 'submitDecisionData' in permissions and not(datasetStatus === 'submitted') %}
```
Gweaton, line 54: Was this not working as expected before?

Gweaton (approved):
This looks good to me - a few minor things but this is good to go after that!

src/helpers/format-status.helper.spec.ts
richpjames: Yeah looking back I think this test is redundant and removing this change is the best option.

views/admin/decisions/show.njk
richpjames: This was formatted automatically on save by Prettier. I opted to go with it as it doesn't seem to make a difference semantically and it's easier to not be fighting Prettier but I am happy to back it out if you'd prefer

src/decisions/presenters/decision-dataset.presenter.spec.ts
richpjames: Yep

views/admin/decisions/show.njk
Gweaton: I'm not sure how reliable Prettier is with `njk` files (we've been struggling to find a reliable config), but it might be better to commit things like this in their own commit (e.e. `Format file with prettier`) so it doesn't introduce overhead in another commit. I don't mind this going in now though.

richpjames: Good point - in future I will add formatting in a seperate commit. 
I am going to re-add the space so it's consistent with the rest of the codebase

PR #642, opened 2022-05-05
(2353) Confirmation pages when editing a dataset

richpjames:
# Changes in this PR
This PR adds a confirmation page letting the user know that if they decided to continue and edit the data there won't be any decision data available until the new draft is published.

I am unsure of a couple of things here: 

- is adding a new controller method for the page the right approach? if so I think the name could do with some work as it doesn't fit in with our very concise controller method names - however I couldn't think of a more appropriate name
- I have opted to pass the 'action' ('unpublish'/'unsubmit') into the translation through the controller and template. I have done this so I can test it in the unit tests as opposed to having to test higher up in e2e tests or add a new test suite. I am very open to feedback as to whether this could be improved. The name 'action' is quite vague but again I was at a loss as to a better name
- I couldn't find a better way to force a linebreak in the copy other than adding two seperate properties for the translation file and manually adding a `<br/>` to the template - is there a better approach?

### Before
The user is taken straight through to edit the data
image

### After
image


bryangaledxw (changes requested):
Looks good. On the points in the PR description

- Maybe follow the `SubmissionController` and `PublicationController` conventions. Follow those, I think the method name would be `new`. Might be worth following the same conventions as those for the locations of the njk files `EditController` uses

- I suggest just pass in the boolean to the template, and use that to choose either `decisions.admin.edit.confirmation.body.unpublish` or `decisions.admin.edit.confirmation.body.unsubmit`, and accept the publication in the i18n file. As it is, we're relying on "unpublish" or "unsubmit" not needing translation

views/admin/decisions/edit/new.njk
```diff
@@ -0,0 +1,31 @@
+{% extends "admin/base.njk" %}
+
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+{% set action = 'unpublished' if datasetPublished else
```
bryangaledxw, line 5: I'd still prefer to see this come from the localisation file - either have `bodyUnpublish` and `bodyUnsubmit`, or lookup the single word from localisation and insert it as you're doing now.

src/decisions/admin/interfaces/edit/new-template.interface.ts
```diff
@@ -0,0 +1,5 @@
+import { EditTemplate } from '../edit-template.interface';
+
+export interface NewTemplate extends Omit<EditTemplate, 'routes'> {
```
bryangaledxw, line 3: I feel like `NewTemplate` and `EditTemplate` aren't fundamentally related - I could see `EditTemplate` changing without wanting `NewTemplate` to.

I feel like either accept the duplication, or rename `PublicationTemplate` to `DatasetDetailsTemplate`, and extend _that_ instead.

Then we end up with three places using `DatasetDetailsTemplate`, with the three places being meaningfully related because they're all sending values into `dataset-details.njk`

bryangaledxw (commented):
Looks better, just a couple of things I want to push back on

src/decisions/admin/interfaces/edit/new-template.interface.ts
bryangaledxw: Also with this being in it's own folder, I'd suggest put `edit-template` with it in the same folder

bryangaledxw (approved):
Good stuff! Would just add the missing newline at the end of `new.njk` before pushing

This is definitely a subjective thing, and I wouldn't worry about it for this PR, but my instinct is to group things so the `foo.njk` file goes in the same commit as the `foo-template.interface.ts` file. My thinking is that these are closely tied, but need some manual verification - they need to match, but if they don't match, that won't get caught by the compiler, so it's good to be able to view them in the same commit

richpjames:
> Good stuff! Would just add the missing newline at the end of `new.njk` before pushing
> 
> This is definitely a subjective thing, and I wouldn't worry about it for this PR, but my instinct is to group things so the `foo.njk` file goes in the same commit as the `foo-template.interface.ts` file. My thinking is that these are closely tied, but need some manual verification - they need to match, but if they don't match, that won't get caught by the compiler, so it's good to be able to view them in the same commit

Thanks for taking another look and your patience with this PR 🙏
Yeah I think you are right - it makes sense to group the interface and the file that uses the interface in one commit - it wasn't a big change so I opted to do it. Thanks for the feedback on that - it's much appreciated... 
Now to fix the merge conflicts  !

PR #643, opened 2022-05-05
(2378) Track search query params and outbound link clicks in Plausible

Gweaton:
# Changes in this PR

Tracks search query params and outbound link clicks in Plausible, so we can get metrics on how effective our search is.

We've had to rollback our change to use the Plausible npm library here to get this working - it may be possible to use the library for this, but time was of the essence.

This isn't the prettiest - industries are logged with their uuid, so we'll need to look these up, but it at least gives us visibility over what users are searching for.

#### Query params showing in Plausible

image

#### Outbound links being tracked

image


richpjames (approved):
Looks good!

PR #646, opened 2022-05-05
(2320) Be able to view decision data for each year as member of the public

bryangaledxw:
# Changes in this PR

- Public view of decision data, linked to from public profession pages

### Before
localhost_3000_professions_registered-trademark-attorney

### After
localhost_3000_professions_registered-trademark-attorney (1)

localhost_3000_decisions_registered-trademark-attorney_2020

### DIFF
--- a/views/professions/_profession-details.njk
+++ b/views/professions/_profession-details.njk
@@ -286,5 +286,12 @@
     ]
   }) }}
 {% endfor %}
+{% endif %}
 
+{% if decisionYears %}
+
+  <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
+  <h2 class="govuk-heading-l">{{ "professions.show.decisions.heading" | t}}</h2>
+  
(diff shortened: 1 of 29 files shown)


views/professions/_profession-details.njk
```diff
@@ -286,5 +286,12 @@
     ]
   }) }}
 {% endfor %}
+{% endif %}
+
+{% if decisionYears %}
```
Gweaton, line 291: I was wondering if we want to hide this until we've deployed the admin-facing link to the dashboard for adding new decision data, but it's probably safe, as there shouldn't be any data, right?

Gweaton (approved):
Looks good to me! Just a couple of small questions/suggestions but nothing blocking

views/professions/_profession-details.njk
bryangaledxw: Yeah, that was my thinking. Conversation for Monday - once this is in, are we ready to add the link to the dashboard?

PR #651, opened 2022-05-09
(2332) Decision data route validations

Gweaton:
# Changes in this PR

Adds validations to display error messages to users when they try to save or submit duplicate or empty routes in decision data.

We can't use the standard validation mechanism here where we use the`@Validate` decorator in the DTO itself due to the shape of the data structure with nested arrays, which make it near-impossible to select the invalid field and highlight it in the table. This adds a custom `DecisionDataValidator` class that wrangles things into the right shape so we can use the existing `ValidationFailedError` class. We highlight the missing fields in the table.

For the controller tests here, I've opted not to mock the error classes for a few reasons:

- We're using other controller methods to test validator logic (even if that's the only way possible using decorator methods)
- I don't want to have loads of slow end to end tests to test this logic
- In an alternative universe, our controller tests would be more integration-style request specs ([see what these look like in Rails](https://relishapp.com/rspec/rspec-rails/docs/request-specs/request-spec) that actually test integrations between the different classes here. Nest calls these "end to end" tests. We've taken to a bit of a hybrid approach here as we've gone with the Nest JS documentation which suggests controllers should be tested as individual units, but I think integration tests would be safer overall.

### After

image

### DIFF
--- a/cypress/integration/admin/decisions/new.spec.ts
+++ b/cypress/integration/admin/decisions/new.spec.ts
@@ -101,6 +101,103 @@ describe('Creating a decision dataset', () => {
           });
         });
     });
+
+    it('displays validation errors when entering invalid data', () => {
+      cy.get('tbody').should(
+        'not.contain',
+        'Profession with no optional fields',
+      );
+
+      cy.translate('decisions.admin.dashboard.addButtonLabel').then((add) => {
+        cy.get('a').contains(add).click();
+      });
+
+      cy.checkAccessibility();
+
+      cy.translate('decisions.admin.new.labels.profession').then(
+        (profession) => {
+          cy.get('label')
+            .contains(profession)
+            .parent()
+            .within(() => {
+              cy.get('select').select('Profession with no optional fields');
+            });
+        },
+      );
+
+      cy.translate('decisions.admin.new.labels.organisation').then(
+        (profession) => {
+          cy.get('label')
+            .contains(profession)
+            .parent()
+            .within(() => {
+              cy.get('select').select('Organisation with no optional fields');
+            });
+        },
+      );
+
+      cy.translate('decisions.admin.new.labels.year').then((profession) => {
+        cy.get('label')
+          .contains(profession)
+          .parent()
+          .within(() => {
+            cy.get('select').select('2021');
+          });
+      });
+
+      cy.translate('app.continue').then((continueLabel) => {
+        cy.get('button').contains(continueLabel).click();
+      });
+
+      cy.get('h1').should('contain', 'Profession with no optional fields');
+
+      cy.checkSummaryListRowValue(
+        'decisions.admin.edit.regulator',
+        'Organisation with no optional fields',
+      );
+      cy.checkSummaryListRowValue('decisions.admin.edit.year', '2021');
+
+      cy.get('input[name="routes[1]"]').type('Duplicate Route');
+      cy.get('select[name="countries[1][1]"]').select('Japan');
+      cy.get('input[name="yeses[1][1]"]').type('3');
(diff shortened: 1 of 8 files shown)


src/decisions/admin/edit.controller.ts
```diff
@@ -112,7 +113,7 @@ export class EditController {
     @Param('professionId') professionId: string,
     @Param('organisationId') organisationId: string,
     @Param('year', ParseIntPipe) year: number,
-    @Body() editDto: EditDto,
```
bryangaledxw, line 115: We can keep the type here, right? I think the only reason we don't have the type when we're doing similar things in other controllers, is to stop validation-via-decorators kicking in because of the default validation pipe

src/decisions/admin/edit.controller.ts
```diff
@@ -130,53 +131,79 @@ export class EditController {
...
+        (unvalidatedActions) => action.includes(unvalidatedActions),
+      )
+    ) {
+      let errors = {};
+      const validator = DecisionDataValidator.validate(editDto);
+
+      if (!validator.valid()) {
+        errors = {
+          ...errors,
+          ...new ValidationFailedError(validator.errors).fullMessages(),
```
bryangaledxw, line 148: I get where this is coming from, but since we've only got the one source of errors, I'd probably go with just `const errors = new new ValidationFailedError(validator.errors).fullMessages()`

src/decisions/admin/edit.controller.ts
```diff
@@ -130,53 +131,79 @@ export class EditController {
...
 
-    const routes = parseEditDtoDecisionRoutes(editDto);
-
     const action = editDto.action;
+    const routes = parseEditDtoDecisionRoutes(editDto);
 
-    if (action === 'publish' || action === 'save' || action === 'submit') {
-      if (action === 'publish') {
-        checkCanPublishDataset(request);
+    if (
```
bryangaledxw, line 137: Why not just stick this entire block inside the `if (action === 'publish' || action === 'save' || action === 'submit')` below? Is the thinking it's safer to check against the actions we don't validate than the ones we do?

src/helpers/decision-data-validator.spec.ts
```diff
@@ -53,4 +53,51 @@ describe('DecisionDataValidator', () => {
...
+        };
+
+        const validated = DecisionDataValidator.validate(editDto);
+
+        expect(validated.valid()).toEqual(true);
+        expect(validated.errors).toEqual([]);
+      });
+    });
+
+    describe('when empty routes are submitted', () => {
```
bryangaledxw, line 77: Should "empty" be "duplicate"?

cypress/integration/admin/decisions/new.spec.ts
```diff
@@ -101,6 +101,103 @@ describe('Creating a decision dataset', () => {
           });
         });
     });
+
+    it.only('displays validation errors when entering invalid data', () => {
```
bryangaledxw, line 105: Slipped in a `.only`

bryangaledxw (approved):
Couple of minor subjective quibbles, but looks good!

Worth keeping a ticket open, as there's gonna be more validation we can do? The ones that come to mind are no plank countries, and no repeated countries in a route

src/decisions/admin/edit.controller.ts
Gweaton: Ah yeah, I initially thought we'd validate when clicking "add country" so that's why this was written the way it was. I decided that probably wasn't so important but didn't think to update it here. thanks!

Gweaton:
@bryangaledxw thanks for the review on this. Have made a few changes.

I've added some tickets to the card for further validation.

PR #653, opened 2022-05-09
(2361) Add backlinks to decision data journey

richpjames:
# Changes in this PR

This PR adds a back link from the decision data edit page to the previous page - the edit confirmation page. 

I checked through the rest of the decision data journey and it doesn't appear that we are missing any other backlinks but please shout up if you think I've missed one.

As mentioned in the commit I wasn't sure whether this should link back to the confirmation page or to the page before that which displays the decision data. I've opted for the confirmation page as from my perspective this would be least confusing for the user. I would happily link it to the page before the confirmation page which just displays the data though if anyone thinks that is more appropriate. 

I also wasn't sure if this should be tested in the e2e tests or if we have full confidence in the BackLink decorator working but opted to be safer than sorry at the cost of a fair amount of back and forth in the e2e tests.

### Before
image

### After
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add confirmation page when editing decision data
 - Validations to ensure no duplicate or empty routes are entered in Decision data
 - Link to recognition decisions dashboard from admin dashboard and navigation bar
+- Add backlink from decision data edit page to the edit confirmation page
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


Gweaton (approved):
I think you're right for the link destination here, the [GOV UK design system says](https://design-system.service.gov.uk/components/back-link/):

> Make sure the link takes users to the previous page they were on, in the state they last saw it.

richpjames:
> I think you're right for the link destination here, the [GOV UK design system says](https://design-system.service.gov.uk/components/back-link/):
> 
> > Make sure the link takes users to the previous page they were on, in the state they last saw it.



> I think you're right for the link destination here, the [GOV UK design system says](https://design-system.service.gov.uk/components/back-link/):
> 
> > Make sure the link takes users to the previous page they were on, in the state they last saw it.

That is good to know! I thought it was best practice.

PR #656, opened 2022-05-10
(2364) Add script for renaming organisations

bryangaledxw:
# Changes in this PR

- Add migration script for renaming organisation


Gweaton (approved):
This looks good - nice one for figuring it out. I was wondering if we'd want to provide the opportunity for users to enter the slug name itself by hand, but I think using the slug generator is a lot safer.

PR #672, opened 2022-05-11
(2337) Guidance for decision data

richpjames:
# Changes in this PR

Adds link from decisions admin dashboard to guidance on decision outcomes and route

### Before
image

### After
image

image

### DIFF
--- a/views/admin/decisions/index.njk
+++ b/views/admin/decisions/index.njk
@@ -14,6 +14,15 @@
 
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
+      <p class="govuk-body">
+        <a class="govuk-link" href="/admin/decisions/guidance">
+          {{ "decisions.admin.dashboard.guidanceLink" | t }}
+        </a>
+      </p>
+    </div>
+  </div>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
       <div class="rpr-internal-listing__manage-decisions-container">
         <div class="rpr-internal-listing__manage-decisions-container-left">
           <h2 class="govuk-heading-l">{{ "decisions.admin.dashboard.manageHeading" | t }}</h2>
@@ -24,7 +33,7 @@
               text: "decisions.admin.dashboard.addButtonLabel" | t,
               href: "/admin/decisions/new"
             }) }}
-          {% endif %}    
+          {% endif %}
         </div>
       </div>
     </div>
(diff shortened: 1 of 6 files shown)


src/decisions/admin/decisions.controller.ts
```diff
@@ -125,4 +125,16 @@ export class DecisionsController {
...
+
+  @Get('guidance')
+  @Permissions(
+    UserPermission.UploadDecisionData,
+    UserPermission.DownloadDecisionData,
+    UserPermission.ViewDecisionData,
+  )
+  @Render('admin/decisions/guidance')
+  @BackLink('/admin/decisions')
+  guidance() {
```
richpjames, line 137: I couldn't think of a way to make this fit the naming convention 🙈
Would it be better to add the method to `app.controller.ts` like the other informational pages ?

bryangaledxw (approved):
Cheers for sorting this!

I think yep move stuff to `AppController` before merging - I feel like this is like the other pages, it just happens for now to be linked to from the decisions dashboard (and that might change once BEIS take a look)

views/admin/decisions/index.njk
```diff
@@ -12,6 +12,13 @@
     {{ title }}
   </h1>
 
+  <span class="govuk-grid-row">
+    <span class="govuk-grid-column-full">
+      <a class="govuk-link govuk-!-font-size-19" href="/admin/decisions/guidance">
```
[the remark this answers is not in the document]
richpjames: I thought giving it an explicit size rather than `govuk-body` might prevent issues down the line. 
I don't have too much familiarity with the govuk DS so my reasoning could be wrong but hypothetically a change could be made to `govuk-body` that changes how it appears other than `font-size` that we wouldn't want to effect this link so I thought explicitly setting the font size would be clearer. 

That being said we're using `govuk-body` elsewhere for links so I think it makes sense to change it for the sake of consistency.

views/admin/decisions/index.njk
```diff
@@ -12,6 +12,13 @@
     {{ title }}
   </h1>
 
+  <span class="govuk-grid-row">
```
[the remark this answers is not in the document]
richpjames: Yep, will have a go

src/decisions/admin/decisions.controller.ts
richpjames: Yep will do, should I copy over the `Permissions` decorator too? Or are we happy with people being able to access that page without the correct permissions - I would guess it's not too much of an issue as they could do no harm with the info.

bryangaledxw: I wouldn't worry about the permissions - with got the other guidance page unprotected

PR #674, opened 2022-05-12
Pin `webpack-encore` package to 1.8.2

bryangaledxw:
The current latest `webpack-encore` depeneds on `webpack` 5.72.0, but `@nestjs/cli` depends on 5.71.0 and these version appear to clash. This can be resolved my manually including the latest `webpack` directly in our `package.json`, though this produces warnings from `webpack-encore`.

Latest `webpack-encore` also produces warning on our stylesheet and latest GDS stylesheets

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "@nestjs/platform-express": "^8.4.4",
         "@nestjs/typeorm": "^8.0.3",
         "@opensearch-project/opensearch": "^1.1.0",
-        "@symfony/webpack-encore": "^1.8.2",
+        "@symfony/webpack-encore": "1.8.2",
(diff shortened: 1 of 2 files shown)


richpjames (approved):
I think this is the right solution for now. Hopefully `@nestjs/cli` will catch up and `webpack-encore` 2.x.x will mature and we can remove this but this works for now. 
Do you think this is worth adding to the changelog? 
Also do you think, going forward, it would be worth adding cards to the board for these kind of changes? (IE dependabot updates that need some kind of manual intervention) so there is visibility of what maintainence work is being done

bryangaledxw:
My feeling is no on the Changelog, it's very much not user-visible. though I've put a ticket in DevOps to unpin this

PR #677, opened 2022-05-12
(2358) nations showing as ‘N/A’ on organisation search results

richpjames:
# Changes in this PR
This seemed to be being caused by the `distinctOn` function in the `allLive` function that was initially added to prevent duplicate orgs showing. 

### Before
image

### After
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Consistent styling on links across service
 - Improve text on internal links to public pages
 - Change "View" buttons to "View details" links on user listing page
+- Prevent organisation's nation from erroneously being shown as 'N/A'
 
 ## [release-021] - 2022-05-11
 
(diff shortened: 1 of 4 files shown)


bryangaledxw (approved):
This makes my very nervous as I can see commit `0b7735c1` specifically added this in 😅

If you've interrogated the ways the duplicates might occur - multiple professions to an org? those professions having multiple versions? - I'm good for this to go in. And definitely check on staging/live once this reaches those

A possible "while you're here" thing (feel free to say this is scope creep) - would it make sense to pass in ProfessionVersionStatus.Live to the `versionsWithJoins` here?

richpjames:
> This makes my very nervous as I can see commit `0b7735c1` specifically added this in 😅
> 
> If you've interrogated the ways the duplicates might occur - multiple professions to an org? those professions having multiple versions? - I'm good for this to go in. And definitely check on staging/live once this reaches those
> 
> A possible "while you're here" thing (feel free to say this is scope creep) - would it make sense to pass in ProfessionVersionStatus.Live to the `versionsWithJoins` here?

Yeah I agree, I'm going to spend some time this afternoon trying to get duplicates showing locally, I was hoping there would be a failing test if they do show but I don't think there is.

bryangaledxw (approved):

PR #697, opened 2022-05-18
(2400) x professions found should be higher up the page for screenreaders

bryangaledxw:
# Changes in this PR

- Move public search result count further up screen for screen readers

Have verified behaviour with macOS VoiceOver


richpjames:
I think this is a good approach, better than the one we spoke about earlier

richpjames (approved):

PR #699, opened 2022-05-18
(2391) - Put multiple regs on new lines in the edit professions list

richpjames:
# Changes in this PR
If a profession has mutltiple organisations in the `/admin/professions` view they will now be comma seperated *and* on a new line

### Before
image

### After
image

### DIFF
--- a/src/professions/admin/presenters/list-entry.presenter.spec.ts
+++ b/src/professions/admin/presenters/list-entry.presenter.spec.ts
@@ -62,7 +62,15 @@ describe('ListEntryPresenter', () => {
 
         const expected: TableRow = [
           { text: 'Example Profession' },
-          { text: 'Example Organisation, Additional Example Organisation' },
+          {
+            html: `<ul class="govuk-list">
+        <li>
+              Example Organisation,
+            </li><li>
+              Additional Example Organisation
+            </li>
+          </ul>`,
+          },
           {
             text: `${translationOf('nations.scotland')}, ${translationOf(
               'nations.northernIreland',
--- a/src/professions/admin/presenters/list-entry.presenter.ts
+++ b/src/professions/admin/presenters/list-entry.presenter.ts
@@ -91,9 +91,24 @@ export class ListEntryPresenter {
       { args: { name: escape(this.profession.name) } },
     )}</a>`;
 
-    const organisations = getOrganisationsFromProfession(this.profession)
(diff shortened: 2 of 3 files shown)


src/professions/admin/presenters/list-entry.presenter.spec.ts
```diff
@@ -62,7 +62,7 @@ describe('ListEntryPresenter', () => {
 
         const expected: TableRow = [
           { text: 'Example Profession' },
-          { text: 'Example Organisation, Additional Example Organisation' },
+          { html: 'Example Organisation,<br/>Additional Example Organisation' },
```
bryangaledxw, line 65: I'm wondering if this should be a list? See `htmlList` in `NationsListPresenter`. As is, this feels a bit like we're adding a third way of presenting things - list (shown as new line separated), comma separated, comma and new line separated

richpjames: Yeah that would be make more sense

src/professions/admin/presenters/list-entry.presenter.ts
```diff
@@ -91,17 +91,32 @@ export class ListEntryPresenter {
       { args: { name: escape(this.profession.name) } },
     )}</a>`;
 
-    const organisations = getOrganisationsFromProfession(this.profession)
-      .map((organisation) => organisation.name)
-      .join(', ');
+    const organisationsList = () => {
```
richpjames, line 94: @bryangaledxw is this more like what you hand in mind?

bryangaledxw: That's it, though I'm still don't think we want the comma? To me this is a 'do for all our lists', or 'do for none of our lists' thing

richpjames: I'm guessing we don't do this anywhere else? 
The reason I left it in is because the line lengths are quite short as it's a table column so there are quite a lot of line breaks and having a comma makes it easier to distingish (at a glance) what is a new list item and what is a regular line break. 
That being said I'm happy to remove it for consistency. It would make the code more readable too :D

bryangaledxw: Ah k, that makes sense, will leave it up to you on this

bryangaledxw (approved):

PR #731, opened 2022-05-24
(2148) Central admin can sort by last updated

bryangaledxw:
# Changes in this PR

- Central users can sort the professions listing

### Before

localhost_3000_admin_professions (1)

### After

localhost_3000_admin_professions


src/professions/admin/presenters/professions.presenter.spec.ts
```diff
@@ -101,7 +101,7 @@ describe('ProfessionsPresenter', () => {
 
       const nations = Nation.all();
 
-      const expected: IndexTemplate = {
+      const expected: Omit<Omit<IndexTemplate, 'filterQuery'>, 'sortMethod'> = {
```
richpjames, line 104: Is this the same as `Omit<IndexTemplate, 'filterQuery' | 'sortMethod'>` ?

assets/stylesheets/application.scss
```diff
@@ -298,6 +298,44 @@ $small-screen: 768px;
       min-height: 2.5em;
       padding: 0 0 0 50px;
     }
+
+    &__sort-container {
+      margin: 1em 0;
+    }
+
+    &__sort-container-left {
```
richpjames, line 306: is this class used?

assets/stylesheets/application.scss
```diff
@@ -298,6 +298,44 @@ $small-screen: 768px;
...
+    &__sort-component-list-item {
+      display: inline-block;
+    }
+
+    &__sort-component-list-item:after {
+      border-right: 1px solid #b1b4b6;
+      content: '';
+      display: inline-block;
+      height: 1.2em;
+      margin-left: 0.5em;
```
richpjames, line 330: I think it would look better if this left & right margin wasn't applied to the last child in the list so the end of the last `sort-component-list-item` lines up with the end of the table/filters.

richpjames (approved):
Great work Bryan!!! 🙌
On the whole it looks great, just some minor comments

src/professions/admin/presenters/professions.presenter.spec.ts
bryangaledxw: Indeed it is, good shout

assets/stylesheets/application.scss
bryangaledxw: Ah, this is applied to the `:after` pseudo-elemt, and is then overrided by: 

```
 &__sort-component-list-item:last-child:after {
      border-right: 0;
}
```

Changing it to:
```
    &__sort-component-list-item:last-child:after {
      border-right: 0;
      margin-left: 0 !important;
      margin-right: 0 !important;
    }
```

Doesn't seem to make any difference? I think any lack of alignment is more down to the contents of the rightmost column being left-aligned, but this text being right-aligned

richpjames: Ah ok, I thought doing what you've done there should have fixed it. 
But as it doesn't I wouldn't worry, was just a minor point 🙂

PR #736, opened 2022-05-26
2406 replace keyword filter on decision page for regulators with their professions

richpjames:
# Changes in this PR


src/decisions/admin/helpers/filter-datasets-on-professions.helper.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export function filterDatasetsOnProfession(
+  datasets: DecisionDataset[],
+  filter: FilterDto,
+): DecisionDataset[] {
+  if (!filter.professions?.length) {
+    return datasets;
+  } else
+    return datasets.filter((dataset) =>
+      filter.professions.includes(dataset.profession.name),
```
bryangaledxw, line 12: Can we go by ID here rather than name?

src/decisions/admin/helpers/get-professions-from-dataset.ts
```diff
@@ -0,0 +1,8 @@
+import { Profession } from '../../../professions/profession.entity';
+import { DecisionDataset } from '../../decision-dataset.entity';
+
+export function getProfessionsFromDataset(
```
bryangaledxw, line 4: Suggest this be `getProfessionsFromDatasets`

src/decisions/admin/presenters/professions-checkbox.presenter.ts
```diff
@@ -0,0 +1,31 @@
+import { CheckboxItems } from '../../../common/interfaces/checkbox-items.interface';
+import { Profession } from '../../../professions/profession.entity';
+
+export class ProfessionsCheckboxPresenter {
+  constructor(
+    private readonly professions: Profession[],
+    private readonly checkedProfessionNames?: string[],
+  ) {}
+
+  private get singleProfessionFromMultipleDatasets(): Profession[] {
```
bryangaledxw, line 10: I feel like this would better fit with `getProfessionsFromDatasets`

src/decisions/admin/presenters/professions-checkbox.presenter.ts
```diff
@@ -0,0 +1,31 @@
...
+    return this.professions.reduce((prevValue, current) => {
+      if (!prevValue.some((profession) => profession.name === current.name))
+        prevValue.push(current);
+      return prevValue;
+    }, []);
+  }
+
+  checkboxItems(): CheckboxItems[] {
+    return this.singleProfessionFromMultipleDatasets?.map((profession) => ({
+      value: profession.name,
```
bryangaledxw, line 20: Can we use `id` rather than `name` here and for `checked`?

views/shared/_filter-criteria.njk
```diff
@@ -49,3 +49,11 @@
     {% endfor %}
   </p>
 {% endif %}
+
+{% if filters.professions %}
```
bryangaledxw, line 53: Can this be `{% if filters.professions | length %}` to match the above blocks? The idea is that an empty array should evaluate to false

views/shared/_filter.njk
```diff
@@ -4,8 +4,27 @@
 
 <form action="" method="get" class="form">
   <div class="{{ filterRowClass }}">
-    <div class="{{ filterColumnClass }}">
-      {{ govukInput({
+
+    {% if showProfessionsFilter %}
```
bryangaledxw, line 8: Can this be `{% if professionsCheckboxItems %}`, then in `index.njk` do:

```
{% set organisationsCheckboxItems = false %}
```
if `view === 'overview'` - we're already doing similar for `organisationsCheckboxItems`

bryangaledxw: Also, can you update the column counter in `_horizontal-filter.njk`

bryangaledxw (commented):
I left a few comments - mostly around keeping things consistent with existing filters, and preferring to go by `id` rather than `name` for identity checking.  I'm afraid I then realised that I don't think the approach of getting the professions list as a "freebie" works. Sorry!

It means that the professions list you see is going to be dependent on what you're doing with the other filters. It might be you see 3 professions, then you filter by year and only see one, and then to get the full list of 3 professions back you first have to uncheck the year and filter again

I don't think you can get away from needing to do another DB call in the `index` method to get the professions list, but that will mean you can just do the filtering inside `applyFilter` in `DecisionDatasetService`, and you won't need to duplicate the filtering between the `index` and `export` methods`

I also wonder if it'll mean you can do `profession: Profession` on `FilterInput` rather than `profession: string`, to keep things strongly typed

src/helpers/create-filter-input.helper.ts
```diff
@@ -29,6 +31,12 @@ export function createFilterInput(
       )
     : undefined;
 
+  const professions = filter.allProfessions
```
bryangaledxw, line 34: Suggest worth expanding the `createFilterInput` tests to cover this

src/decisions/admin/decisions.controller.ts
```diff
@@ -62,9 +65,19 @@ export class DecisionsController {
 
     const allOrganisations = await this.organisationVersionsService.allLive();
 
+    let allProfessions: Profession[] = [];
```
bryangaledxw, line 68: Definitely a subjective thing, but my preference is generally to go for `const` where we can for the benefit of being able to look at a value and know you don't need to worry about it changing, so we'd do

```
const allProfessions = userOrganisation ? await ... : []
```

src/decisions/admin/presenters/professions-checkbox.presenter.ts
```diff
@@ -0,0 +1,23 @@
+import { CheckboxItems } from '../../../common/interfaces/checkbox-items.interface';
+import { Profession } from '../../../professions/profession.entity';
+
+export class ProfessionsCheckboxPresenter {
+  constructor(
+    private readonly professions: Profession[],
+    private readonly checkedProfessionNames?: Profession[],
```
bryangaledxw, line 7: This can just be `checkedProfessions`

src/decisions/admin/interfaces/index-template.interface.ts
```diff
@@ -14,11 +15,12 @@ export interface IndexTemplate {
...
     years: number[];
     statuses: string[];
+    professions?: Profession[];
   };
 
   organisationsCheckboxItems: CheckboxItems[];
   yearsCheckboxItems: CheckboxItems[];
   statusesCheckboxItems: CheckboxItems[];
-
+  professionsCheckboxItems?: CheckboxItems[];
```
bryangaledxw, line 24: Does this need to be `?` I think something always gets set for this

src/decisions/admin/presenters/decision-datasets.presenter.ts
```diff
@@ -20,6 +20,7 @@ export class DecisionDatasetsPresenter {
     private readonly endYear: number,
     private readonly decisionDatasets: DecisionDataset[],
     private readonly i18nService: I18nService,
+    private readonly allProfessions?: Profession[],
```
bryangaledxw, line 25: Similar to `IndexTemplate`, does this need to be `?`. And suggest move this up a line, so with got all the "actual" input, then the `I18nService`

views/admin/decisions/index.njk
```diff
@@ -44,6 +44,8 @@
       <p class="govuk-body">{{"decisions.admin.dashboard.instructions" | t}}</p>
       {% if view !== 'overview' %}
         {% set organisationsCheckboxItems = false %}
+      {% elif view === 'overview' %}
+        {% set professionsCheckboxItems = false %}
       {% endif %}
```
bryangaledxw, line 49: Suggest rearrange to get rid of the redundancy - something like
```
{% if view === 'overview' %}
  ...
{% elif %}
  ...
{% endif %}
```

bryangaledxw (approved):
Good stuff! A few minor comments, but nothing approval-blocking

PR #741, opened 2022-05-26
(2399) Left align the decision data download button

richpjames:
# Changes in this PR

Left align the decision data download button and add some padding beneath it so it isn't too close to the 'X Datasets found' heading

### Before
image
### After
image


richpjames:
@bryangaledxw the css properties are a lot closer to the ones on the prototype now. 
I have had to keep the `margin-bottom` on the `__file-download-link` class otherwise the `govuk-body` adds too much margin on the bottom but hopefully that is acceptable. 
I have had a look at the individual dataset page and the download link appears uneffected

bryangaledxw (approved):

PR #751, opened 2022-05-31
(2324) Everywhere we use `translate` no longer needs to return a promise

bryangaledxw:
# Changes in this PR

- Remove unnecessary `awaits` from use of `I18nService.translate`


richpjames (approved):
Well done! You deserve a medal for this 🏅

PR #758, opened 2022-05-31
(2300) Allow archived organisations to be unarchived

richpjames:
# Changes in this PR
This is a similar journey to archiving only when an org is unarchived its status is set to 'Live' instead of 'Archived'. 

I believe this necessitates a new method in the `organisation-versions.service` which I've created but with a very naïve implementation, there may be more work needed here?
 

### Before
image

### After
image
image
image

### DIFF
--- a/views/admin/organisations/show.njk
+++ b/views/admin/organisations/show.njk
@@ -35,15 +35,15 @@
         {% if hasLiveVersion %}
           <h2 class="govuk-heading-s" data-cy="currently-published-version-text">
             {{ "organisations.admin.publicFacingLink.heading" | t }}<br>
-            <span class="govuk-body" data-cy="currently-published-version"><a class="govuk-link" href="/regulatory-authorities/{{ organisation.slug }}">{{ "organisations.admin.publicFacingLink.label" | t }}</a></span>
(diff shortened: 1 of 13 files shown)


bryangaledxw (commented):
Good stuff getting this done so quickly!

Biggest thing I think to change is bring the unarchive controller closer to the archive controller. A lot of how we're doing things - keeping more versions around than we really need to - is driven by wanting to keep a complete papertrail of things in the DB so it can theoretically at least be inspected. So "unarchive" is really "create a duplicate draft that supersedes the archived version"

bryangaledxw (approved):
Looks good - just a couple of things to fix before merging -
- You're checking for `publishOrganisation` in the template, but `createOrganisation` in the controller
- You're still referencing the permission being "publish" in the message for the commit where you update the controller

views/admin/organisations/show.njk
```diff
@@ -80,10 +79,20 @@
                 }}
               </li>
               {% endif %}
+              {% if organisation.status === 'archived' and 'publishOrganisation' in permissions %}
```
bryangaledxw, line 82: I think this wants to be `createOrganisation` to match the controller

PR #759, opened 2022-06-01
(2247) Tidy up how we fetch a users organisation

bryangaledxw:
# Changes in this PR

- Consistently use `getUserOrganisation`, rather than having multiple places in code that are aware of the "app.beis" string


bryangaledxw:
Yep, that was the thinking

PR #763, opened 2022-06-01
(2050) Prevent tests running in wrong environment

richpjames:
There have been issues with users getting deleted because e2e tests have been run in the development environment.

This change adds a visually hidden test banner to the top of each page which Cypress checks exists before any test run.

Whilst this will prevent the tests running in a non-test environment there will be a performance penalty for running this assertion before every test

### DIFF
--- a/cypress/support/index.ts
+++ b/cypress/support/index.ts
@@ -103,6 +103,10 @@ import './commands';
 // Purge the Opensearch index and seed the database
 // before running the specs
 before(() => {
+  cy.visit('')
+    .get('[data-cy=environment-marker]')
+    .should('contain', 'test environment');
+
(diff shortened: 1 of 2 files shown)


richpjames:
Unfortunately this change won't work without further work on the test files themselves. My time on this project is finished and although implementing [the suggestion in the test failures (using `cy.origin`)](https://github.com/UKGovernmentBEIS/regulated-professions-register/runs/6752732155?check_suite_focus=true) may work I am reluctant to introudce such a change this late in development as it could cause flakiness in E2E that the future custodians of this project would have to deal with. Further, this additional environment test should be as light touch as possible and having to make changes across multiple test suites doesn't seem to be worth the benefits it would bring at this point in the project lifecycle.

